PR #2, opened 2022-07-01
Auth

BenRamchandani:
More faff than it has any right to be, but now have a hopefully vaguely sensible auth setup.

* To log someone in redirect them to `/login?returnUrl=/something`
* This redirects to Google, they sign in
* Google redirects them back to `/signin-google`, which the auth library magically handles, sets a session cookie, and sends them back to `returnUrl` from above
* Session cookie has their google account id and stuff in, on each request we read that and look them up in our database (making them if they don't exist)
    * We then replace `HttpContext.User.Identity` with a `BricksAndHeartsUser` object that can have the properties we want on
* The User table has an "IsAdmin" column, which gives that user the "Admin" role, so we can do `[Authorize(Roles = "Admin")]`, admins come through, others get a 403, and if you're not logged in it redirects to `/login`


web/appsettings.json
```diff
@@ -6,5 +6,11 @@
...
   },
   "AllowedHosts": "*",
-  "DBConnectionString": "Server=(local);Database=BricksAndHearts;Trusted_Connection=True;Integrated Security=True;"
+  "DBConnectionString": "Server=(local);Database=BricksAndHearts;Trusted_Connection=True;Integrated Security=True;",
+  "Authentication" : {
+    "Google" : {
+      "ClientId" : "",
+      "ClientSecret" : ""
+    }
+  }
```
BenRamchandani, line 15: I added empty placeholders so we can see the shape. I haven't set these in Azure yet, or migrated the Azure database

PR #7, opened 2022-07-18
BNH-2 Request administrator access

Chestnut-lol:


web/Controllers/AdminController.cs
```diff
@@ -26,4 +29,40 @@ public IActionResult Index()
...
+    [HttpPost]
+    public IActionResult RequestAdminAccess()
+    {
+        var isAuthenticated = User.Identity?.IsAuthenticated ?? false;
+
+        if (!isAuthenticated)
+        {
+            _logger.LogWarning("Not authenticated user");
+            TempData["FlashType"] = "danger";
+
```
emilypatsko, line 43: Unnecessary line break - remove to be consistent with the same pattern  you've used below :)

web/Controllers/AdminController.cs
```diff
@@ -26,4 +29,40 @@ public IActionResult Index()
...
+            _logger.LogWarning("User {UserId} already an admin",user.Id);
+            TempData["FlashType"] = "danger";
+            TempData["FlashMessage"] = "Already an admin";
+
+            return RedirectToAction(nameof(Index));
+        }
+
+        _adminService.RequestAdminAccess(user);
+        TempData["FlashType"] = "success";
+
```
emilypatsko, line 61: Remove line break

web/Controllers/AdminController.cs
```diff
@@ -26,4 +29,40 @@ public IActionResult Index()
...
+        TempData["FlashMessage"] = "Successfully requested admin access";
+        _logger.LogInformation("Successfully requested admin access for user {UserId}",user.Id);
+        return RedirectToAction(nameof(Index));
```
emilypatsko, line 64: I'd have a line break before this line 😄

web/Migrations/20220718092020_HasRequestedAdmin.cs
```diff
@@ -0,0 +1,26 @@
+﻿using Microsoft.EntityFrameworkCore.Migrations;
+
+#nullable disable
+
+namespace BricksAndHearts.Migrations
+{
+    public partial class HasRequestedAdmin : Migration
```
emilypatsko, line 7: Possibly a bit pedantic, but I generally like migrations to describe what they do, so would name this something like `AddHasRequestedAdminToUser`

web/Views/Admin/Index.cshtml
```diff
@@ -19,11 +29,17 @@
             else
             {
                 <p>You are not an admin :(</p>
+                @using (Html.BeginForm("RequestAdminAccess", "Admin", FormMethod.Post))
```
emilypatsko, line 32: Could you add a screenshot of what this change looks like? Good practice to do this for any frontend changes

emilypatsko (changes requested):

web/Migrations/20220718092020_HasRequestedAdmin.cs
MaxTheComputerer: Will keep this in mind for future. Is it possible to rename this migration now or is it too late?

emilypatsko: You can undo a migration, delete it, and recreate it with a different name. Something like
```
// Update your database to the latest good migration
dotnet ef database update <name of previous migration>

// Then remove the bad one (the last one)
dotnet ef migrations remove

// Then create the new one
dotnet ef migrations add <new migration name>

// and update
dotnet ef database update
```

You might want to do some googling and check out these commands yourself but I think this should do the trick :)

web/Controllers/AdminController.cs
```diff
@@ -26,4 +29,40 @@ public IActionResult Index()
...
+        if (user.IsAdmin)
+        {
+            _logger.LogWarning("User {UserId} already an admin",user.Id);
+            TempData["FlashType"] = "danger";
+            TempData["FlashMessage"] = "Already an admin";
+
+            return RedirectToAction(nameof(Index));
+        }
+
+        _adminService.RequestAdminAccess(user);
```
emilypatsko, line 58: Is there anything that could go wrong in this method? Any errors to catch?

emilypatsko (changes requested):

web/Controllers/AdminController.cs
BenRamchandani: FWIW I would vote not to have the extra error handling for the user not being found. It's not a clear-cut thing, but I think a generic error page is reasonable - currently it should be impossible, and even if we added a way to delete users that happening between the start of the request and here (we check the user against the database on every request currently) is unlikely and would probably mean there's a bug rather than the user doing something wrong.

Of course things could change and the method might be called in a way that makes it more likely, but I'd generally prefer to keep things simple.

Anyway feel free to keep the code or revert it, fine by me either way.

(it's fine to answer "No" or start a discussion to review comments from us if you don't think a suggestion makes sense, and especially when we're asking a question)

emilypatsko: Ah you beat me to it Ben 😄 I was just going to say, this isn't quite what I had intended. I was thinking more along the lines of "if `GetCurrentUser` returns `null`, do something" but I'm not even sure if that method can return `null`.

Like Ben says, when I comment a question it's not a trick, I'm genuinely starting a discussion 😄 I'd probably revert the code to the way it was, as it was clearer to see what was going on. 😊

emilypatsko (changes requested):
As mentioned in the thread with Ben, I think your `RequestAdminAccess` method was nicer before. I reckon revert it back to that, and then this is probably good to go

web/Controllers/AdminController.cs
MaxTheComputerer: It's been reverted! Sorry about that 😅

emilypatsko (approved):

PR #9, opened 2022-07-18
BNH13

SleepyUser:
Added Error pages for a range of status codes. See ErrorService file for full list.

ErrorView returns returns the Code, and a Name and Message associated with that code (if it is listed, otherwise a generic message is displayed).

Formatting and design of the error page still required in future to bring in line with overall visual style of site.

### DIFF
--- a/web/Program.cs
+++ b/web/Program.cs
@@ -44,7 +44,7 @@ if (app.Configuration.GetValue<bool>("MigrateOnStartup"))
 
 if (!app.Environment.IsDevelopment())
 {
-    app.UseExceptionHandler("/Home/Error");
+    app.UseExceptionHandler("/Home/Error/500");
     // The default HSTS value is 30 days. You may want to change this for production scenarios, see https://aka.ms/aspnetcore-hsts.
     app.UseHsts();
     //app.UseHttpsRedirection();
--- a/web/Program.cs
+++ b/web/Program.cs
@@ -47,8 +47,10 @@ if (!app.Environment.IsDevelopment())
     app.UseExceptionHandler("/Home/Error");
     // The default HSTS value is 30 days. You may want to change this for production scenarios, see https://aka.ms/aspnetcore-hsts.
     app.UseHsts();
-    app.UseHttpsRedirection();
+    //app.UseHttpsRedirection();
 }
+
+app.UseStatusCodePagesWithReExecute("/Error/{0}");
 
 app.UseCookiePolicy(
     new CookiePolicyOptions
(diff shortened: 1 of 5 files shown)


web/Program.cs
```diff
@@ -47,9 +47,11 @@
     app.UseExceptionHandler("/Home/Error");
```
BenRamchandani, line 47: Should change this to "/Home/Error/500"

BenRamchandani (commented):
This approach seems sensible to me. If you're happy with it tidy it up a bit and then ping me again and I can have another look.

web/ViewModels/ErrorViewModel.cs
```diff
@@ -5,4 +5,8 @@ public class ErrorViewModel
     public string? RequestId { get; set; }
 
     public bool ShowRequestId => !string.IsNullOrEmpty(RequestId);
+
+    public string? StatusName { get; set; }
```
emilypatsko, line 9: Do these need to be nullable?

web/Services/ErrorService.cs
```diff
@@ -0,0 +1,46 @@
+﻿namespace BricksAndHearts.Services;
+public class ErrorService
```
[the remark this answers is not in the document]
BenRamchandani: If this is just going to have a static method I'd be fine for it not to have an interface, but we shouldn't call it service, could call it e.g. `ErrorCodeMapping`

web/Views/Shared/Error.cshtml
```diff
@@ -9,11 +9,14 @@
...
-        <strong>Request ID:</strong> <code>@Model.RequestId</code>
+        <strong>Error Code: @Model.RequestId</strong> <code>@Model.StatusName</code>
+    </p>
+    <p>
+        <code>@Model.StatusMessage</code>
     </p>
 }
 
-<h3>Development Mode</h3>
+@*<h3>Development Mode</h3>
```
BenRamchandani, line 19: Let's delete this, we don't need the commented out stuff

web/Services/ErrorService.cs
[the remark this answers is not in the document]
cec200wire: We changed it to use an interface.

web/Services/ErrorService.cs
```diff
@@ -0,0 +1,51 @@
+﻿namespace BricksAndHearts.Services;
+
+public interface IErrorService
+{
+    public (string statusName, string statusMessage) GetStatusMessage(int statusCode);
+}
+public class ErrorService : IErrorService
+{ 
+    public (string statusName, string statusMessage) GetStatusMessage(int statusCode)
```
emilypatsko, line 9: I think these should be in pascal case, i.e. `StatusName` not `statusName`, pedantic I know 😄

emilypatsko (dismissed):
I've just asked for two very small changes in `ErrorService`, please make these but then go ahead and merge :)

emilypatsko (approved):

emilypatsko (approved):

PR #17, opened 2022-07-19
BNH-22 Properties tab

MaxTheComputerer:
- Added `Property` table to database, each of which are associated with a Landlord
- A landlord can view a table of all of their properties
- This can be accessed from a "View Properties" link on the navbar which is visible to logged in users

image


web/Views/Shared/_Layout.cshtml
```diff
@@ -26,6 +26,12 @@
                         <li class="nav-item">
                             <a class="nav-link text-dark" asp-area="" asp-controller="Home" asp-action="Privacy">Privacy</a>
                         </li>
+                        @if (User.Identity?.IsAuthenticated == true)
+                         {
+                             <li class="nav-item">
+                                 <a class="nav-link text-dark" asp-area="" asp-controller="Landlord" asp-action="ViewProperties">View Properties</a>
+                             </li>
+                         }
```
MaxTheComputerer, line 34: We can't find a way to check if a user is a landlord from this `_Layout` file (and I think @TomHanmer29 and @blancamc00 had a similar issue). For now, we show this link to any authenticated user, then if a non-landlord clicks the link they get a 403.

A possible solution would be to add a `Landlord` role, similar to the existing `Admin` one. We could then do something like `@if (User.IsInRole("Landlord"))` to selectively show navbar elements. But I thought this might warrant its own ticket.

MaxTheComputerer: Alternatively, `((BricksAndHeartsUser) User.Identity).LandlordId` seems to work without any changes. Which option would be better here? The role approach might also be useful for `[Authorize(Roles = "Landlord")]` on other pages.

web/Migrations/20220718142048_CreatePropertiesTable.cs
```diff
@@ -0,0 +1,43 @@
...
+                },
+                constraints: table =>
+                {
+                    table.PrimaryKey("PK_Property", x => x.Id);
+                    table.ForeignKey(
+                        name: "FK_Property_Landlord_LandlordId",
+                        column: x => x.LandlordId,
+                        principalTable: "Landlord",
+                        principalColumn: "Id",
+                        onDelete: ReferentialAction.Cascade);
```
BenRamchandani, line 28: I don't think we want this to cascade. Looks like you'll need to configure it in the DB context then recreate the migration: https://docs.microsoft.com/en-us/ef/core/saving/cascade-delete#configuring-cascading-behaviors

web/Migrations/20220718143919_AddListOfPropertiesToLandlord.cs
```diff
@@ -0,0 +1,19 @@
+﻿using Microsoft.EntityFrameworkCore.Migrations;
+
+#nullable disable
+
+namespace BricksAndHearts.Migrations
+{
+    public partial class AddListOfPropertiesToLandlord : Migration
```
BenRamchandani, line 7: Can get rid of the empty migration

web/Services/LandlordService.cs
```diff
@@ -71,4 +73,16 @@ await using (var transaction = await _dbContext.Database.BeginTransactionAsync(I
...
         return ILandlordService.LandlordRegistrationResult.Success;
     }
+
+    public List<PropertyViewModel> GetListOfProperties(int landlordId)
+    {
+
+        var landlord = _dbContext.Landlords
+            .Include(l => l.Properties)
+            .SingleOrDefault(l => l.Id == landlordId);
+        // Assumes landlord does exist
```
BenRamchandani, line 83: Then use `.Single`, not `.SingleOrDefault`. I'd also probably either:

* Change this to query Properties directly
* Or have the method return the landlord (i.e. `GetLandlordWithProperties`)

web/Services/LandlordService.cs
```diff
@@ -71,4 +73,16 @@ await using (var transaction = await _dbContext.Database.BeginTransactionAsync(I
...
+        var listOfProperties = new List<PropertyViewModel>(); 
+        landlord.Properties.ForEach(p=> listOfProperties.Add(PropertyViewModel.FromDbModel(p)));
```
BenRamchandani, line 85: Traditionally the controller handles mapping to the View Model. Obviously it doesn't really matter, but may as well follow standard practice.

web/Views/Shared/_Layout.cshtml
BenRamchandani: Lets do `((BricksAndHeartsUser) User.Identity).LandlordId != null` for now.

Adding a new role sounds sensible, agreed it should be a separate PR at least, and may as well have a ticket, do raise one.

BenRamchandani (changes requested):

BenRamchandani (approved):
Couple of tiny things, but feel free to merge

MaxTheComputerer:
Rebased onto main and resolved conflicts in the navbar

BenRamchandani (approved):

PR #18, opened 2022-07-19
BNH-3 Add button to cancel admin rights request

MaxTheComputerer:
If a user has requested admin access, they can now cancel this request. A confirmation message appears when this is successful.

image

image

### DIFF
--- a/web/Controllers/AdminController.cs
+++ b/web/Controllers/AdminController.cs
@@ -50,4 +50,25 @@ public class AdminController : AbstractController
 
         return RedirectToAction(nameof(Index));
     }
+
+    public IActionResult CancelAdminAccessRequest()
+    {
+        var user = GetCurrentUser();
+        if (user.IsAdmin)
+        {
+            _logger.LogWarning("User {UserId} already an admin",user.Id);
+            TempData["FlashType"] = "danger";
+            TempData["FlashMessage"] = "Already an admin";
+
+            return RedirectToAction(nameof(Index));
+        }
+
+        _adminService.CancelAdminAccessRequest(user);
+
+        _logger.LogInformation("Successfully cancelled admin access request for user {UserId}",user.Id);
+        TempData["FlashType"] = "success";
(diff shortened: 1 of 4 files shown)


web/Services/AdminService.cs
```diff
@@ -23,4 +24,11 @@ public void RequestAdminAccess(BricksAndHeartsUser user)
         userRecord.HasRequestedAdmin = true;
         _dbContext.SaveChanges();
     }
+    
+    public void CancelAdminAccessRequest(BricksAndHeartsUser user)
+    {
+        var userRecord = _dbContext.Users.SingleOrDefault(u => u.Id == user.Id);
```
BenRamchandani, line 30:
```suggestion
        var userRecord = _dbContext.Users.Single(u => u.Id == user.Id);
```

web/Controllers/AdminController.cs
```diff
@@ -50,4 +50,25 @@ public IActionResult RequestAdminAccess()
...
     }
+
+    public IActionResult CancelAdminAccessRequest()
+    {
+        var user = GetCurrentUser();
+        if (user.IsAdmin)
+        {
+            _logger.LogWarning("User {UserId} already an admin",user.Id);
+            TempData["FlashType"] = "danger";
+            TempData["FlashMessage"] = "Already an admin";
```
BenRamchandani, line 61: We should pull this out into a method on AbstractController at some point since we're using it quite a bit. Definitely separate PR though

BenRamchandani (approved):
Looks good. One minor comment then fine to merge. Also there's some really minor formatting things, it's worth getting into the habit of formatting in Rider, either as you go or with the Reformat and Cleanup -> "Uncommitted Code" before you commit (but don't worry for this PR)

web/Controllers/AdminController.cs
MaxTheComputerer: Will make a ticket :)

MaxTheComputerer:
All changes made!

BenRamchandani (approved):
👍

PR #19, opened 2022-07-19
BNH-16 Landlord Sign Up

TomHanmer29:
Updated landlord sign up page to be split across several columns. added buttons to move between pages and added 2 new columns to the database for landlord charter status and landlord type based on the fields in the changeAhead landlord sign up form. 
image
image
user inputs are checked for being empty when you move through the pages of the form, and checked for invalid inputs upon form submission.

### DIFF
--- a/web/wwwroot/js/site.js
+++ b/web/wwwroot/js/site.js
@@ -1,4 +1,68 @@
-﻿// Please see documentation at https://docs.microsoft.com/aspnet/core/client-side/bundling-and-minification
-// for details on configuring this project to bundle and minify static web assets.
+﻿let currentTab = 0;
+showTab(currentTab);
 
-// Write your JavaScript code.
+function showTab(currentTab) {
+    let tabList = document.getElementsByClassName("registerTab");
+    tabList[currentTab].style.display = "block";
+    if (currentTab == 0) {
+        document.getElementById("prevBtn").style.display = "none";
+    } else {
+        document.getElementById("prevBtn").style.display = "inline";
+    }
+    if (currentTab == (tabList.length - 1)) {
+        document.getElementById("nextBtn").innerHTML = "Submit";
+    } else {
+        document.getElementById("nextBtn").innerHTML = "Next";
+    }
+    fixStepIndicator(currentTab)
+}
+
+function nextPrev(tabChange) {
+    let tabList = document.getElementsByClassName("registerTab");
+    // Exit the function if any field in the current tab is invalid:
+    if (tabChange == 1 && !validateForm()) return false;
+    // Hide the current tab:
+    tabList[currentTab].style.display = "none";
+    // Increase or decrease the current tab by 1:
+    currentTab = currentTab + tabChange;
+    // if you have reached the end of the form... :
+    if (currentTab >= tabList.length) {
+        //...the form gets submitted:
+        document.getElementById("landlordEntry").submit();
+        return false;
+    }
+    // Otherwise, display the correct tab:
+    showTab(currentTab);
+}
+
+function validateForm() {
+    // This function deals with validation of the form fields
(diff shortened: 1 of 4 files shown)


TomHanmer29:
I think my commit history is a bit messy - I tried to clean it up a bit but not really sure it worked. At some point if someone could give me a hand with how to tidy up my commit history that would be very appreciated :)

TomHanmer29:
I think I've realised I ended up pulling remote when my branch was behind remote and should have maybe just force pushed (based on chat with ben abt one of Blanca's branches)

web/Migrations/20220719095559_AddCharterAndLandlordStatusToLandlordAddHasRequestedAdminToUser.cs
```diff
@@ -0,0 +1,48 @@
...
+    public partial class AddCharterAndLandlordStatusToLandlordAddHasRequestedAdminToUser : Migration
+    {
+        protected override void Up(MigrationBuilder migrationBuilder)
+        {
+            migrationBuilder.AddColumn<bool>(
+                name: "HasRequestedAdmin",
+                table: "User",
+                type: "bit",
+                nullable: false,
+                defaultValue: false);
```
BenRamchandani, line 16: This shouldn't be here.. make sure you're rebased onto main and try recreating the migration

web/Database/LandlordDbModel.cs
```diff
@@ -13,5 +13,8 @@ public class LandlordDbModel
     public string Email { get; set; } = null!;
     public string Phone { get; set; } = null!;
 
+    public string LandlordStatus { get; set; } = null!;
+    public bool CharterStatus { get; set; } = false;
```
BenRamchandani, line 17: If this is provided by the landlord then we should make that explicit so it's not confused with the admin approval for charter. I'd call it e.g. LandlordProvidedCharterStatus

web/wwwroot/js/site.js
```diff
@@ -1,4 +1,68 @@
...
+        document.getElementsByClassName("step")[currentTab].className += " finish";
+    }
+    return valid;
+}
+
+function fixStepIndicator(currentStep) {
+    // This function removes the "active" class of all steps...
+    let stepList = document.getElementsByClassName("step");
+    for (let stepIterator = 0; stepIterator < stepList.length; stepIterator++) {
+        stepList[stepIterator].className = stepList[stepIterator].className.replace(" active", "");
```
BenRamchandani, line 64: Same as this I think (though I haven't tested it!)
```suggestion
        stepList[stepIterator].classList.remove("active");

```

BenRamchandani (commented):
Ohh pretty

I've put a couple of comments on. For git I think the approach I'd take is:

* Check out branch
* Soft reset to the point it branched off main (this is not necessarily the same as current main)
    * This kills all the commits, but leaves your working directory
* Add the db model and migration changes and put them in one commit
* Put the rest in another commit
* Check it all works and force push
* Rebase onto main if necessary

I'm happy to help or you can give it a go yourself

web/Migrations/20220719095559_AddCharterAndLandlordStatusToLandlordAddHasRequestedAdminToUser.cs
BenRamchandani: It looks like the existing migration 20220718155426_AddHasRequestedAdminToUser is deleted on this branch. Can you bring that back, then regenerate this migration, which then shouldn't be trying to add this column again

BenRamchandani (changes requested):
Need to sort out the migrations, but otherwise this looks good to me.

From commit message
> added 2 new rows to landlord in database
should be "columns" not rows

PR #20, opened 2022-07-19
BNH-4 - Admin List

SleepyUser:
Added functionality to allow all pending (but not current) admins to show in a pending admins list.
Added functionality to allow all admins to show in a list.
Only allows the list to show if user is an admin.

Added a link from the admin index to the Admin List.


cec200wire:
image
(Admin list display)

web/Views/Admin/AdminList.cshtml
```diff
@@ -0,0 +1,62 @@
...
+    ViewData["Title"] = "Admin List";
+}
+
+<!DOCTYPE html>
+
+<html lang="en">
+    <head>
+        <title>Admin List</title>
+    </head>
+<body>
```
BenRamchandani, line 14: It should be using `_Layout` so you shouldn't need this preamble

BenRamchandani (commented):
Couple of comments, but looks sensible.

Should we add a link to the nav bar if you're an admin? Feels like we may as well.

cec200wire:
Sorry, I had to rebase again and this caused a minor error (one of the async things lost the Task<> label) - I fixed this and checked things still work, but I now can't merge without another approval.

PR #24, opened 2022-07-20
BNH-23 Create property

MaxTheComputerer:
- Landlords can now add a new property that will be linked to themselves
- Properties have more fields available to set now, such as property type, number of bedrooms, rent, etc.
- A property's address has also been expanded into multiple fields
- A `PropertyService` has been created for property-related database operations
- View Properties table still needs styling properly

image


web/ViewModels/PropertyViewModel.cs
```diff
@@ -1,17 +1,56 @@
...
 using BricksAndHearts.Database;
 
 namespace BricksAndHearts.ViewModels;
 
 public class PropertyViewModel
 {
-    [Required] public string Address { get; set; } = string.Empty;
+    [Required] public PropertyAddress Address { get; set; }
+
+    [Required] public string PropertyType { get; set; } = string.Empty;
```
BenRamchandani, line 11: I haven't generally been asking for this, but it's good practice to put sensible limits on field size by default, e.g. `[StringLength(10000)]`. They should be much higher than could ever be reasonable to actually use.

It's not a big deal, but it makes it harder to do denial of service attacks, e.g. no-one needs to be able to add a 100 million character address line and that might just break the website.

BenRamchandani (changes requested):
Nice work. Left a few comments, though there's less to comment on so I'm getting more pedantic.

You'll also need to rebase to get the workflow file so it can run the tests (and adding a couple of tests wouldn't hurt)

web/Services/PropertyService.cs
```diff
@@ -0,0 +1,67 @@
...
+            await _dbContext.SaveChangesAsync();
+
+            // Now we can update the landlord record too
+            var landlordRecord = _dbContext.Landlords.Single(l => l.Id == landlordId);
+            landlordRecord.Properties.Add(dbModel);
+            await _dbContext.SaveChangesAsync();
+
+            // and finally commit
+            await transaction.CommitAsync();
+        }
```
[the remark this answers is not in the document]
MaxTheComputerer: Ah that makes sense. I'll remove the transaction then! Out of interest, under what circumstances is it useful to use `async`? It seems most of our database operations in the project so far don't use it, but `RegisterLandlordWithUser` does.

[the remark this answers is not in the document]
BenRamchandani: Generally adding `async` on anything that's going to take more than a handful of microseconds will be a performance improvement as request load increases - `async` things run inside an event loop which will context switch much faster than OS threads which you need if you're using blocking IO.

That said it doesn't really matter for an app like this, we're unlikely to ever need to support >1000 concurrent users, so I haven't been pushing it much in reviews just to keep things simple.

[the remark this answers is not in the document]
MaxTheComputerer: That's good to know, thank you!

MaxTheComputerer:
Added unit tests and rebased onto main

web/ViewModels/PropertyViewModel.cs
```diff
@@ -6,22 +6,22 @@ namespace BricksAndHearts.ViewModels;
 
 public class PropertyViewModel
 {
-    [Required] public PropertyAddress Address { get; set; }
+    [Required] public PropertyAddress? Address { get; set; }
```
kwunkiuma, line 9: Can we try to address some of these warnings with default values as opposed to making them nullable? Especially this Address field so that we can get rid of the non-null assertions in the view.

kwunkiuma: You can set this to a default Address object with `= new();` 🙂

kwunkiuma (changes requested):
Looks good, just some suggestions for how warnings are addressed 🙂

web/ViewModels/PropertyViewModel.cs
MaxTheComputerer: Done!

web/ViewModels/PropertyViewModel.cs
```diff
@@ -49,9 +49,9 @@ public class PropertyAddress
 {
     [Required] [StringLength(10000)] public string AddressLine1 { get; set; } = string.Empty;
 
-    [StringLength(10000)] public string? AddressLine2 { get; set; }
+    [StringLength(10000)] public string AddressLine2 { get; set; } = string.Empty;
```
kwunkiuma, line 52: Ah sorry I should have been more clear 😅 it's not an issue for fields that we expect can be null to be nullable, so it's alright for optional fields like AddressLine2/3 which are nullable in the database to be nullable here as well 🙂

kwunkiuma (changes requested):
Looks good, just one small thing!

web/ViewModels/PropertyViewModel.cs
MaxTheComputerer: Ah I see! They'll almost all become nullable in BNH-34 either way

web/Views/Landlord/Properties.cshtml
```diff
@@ -7,23 +7,35 @@
...
+    </thead>
+    <tbody>
+    @{
+        var i = 1;
+    }
+    @foreach (var prop in Model.Properties)
+    {
+        <tr>
+            <th scope="row">@i</th>
+            <td>@prop.Address!.AddressLine1, @prop.Address.AddressLine2, @prop.Address.AddressLine3, @prop.Address.TownOrCity</td>
```
kwunkiuma, line 29: We can add some null checks so that we don't display "Line 1, , , City" when address lines 2 and 3 are empty. You might update how the values are stored so that they save as null as opposed to empty string when the input textbox is empty.

Do let me know if this is something better addressed in BNH-34, then we can get this merged first 🙂

web/ViewModels/PropertyViewModel.cs
```diff
@@ -1,17 +1,61 @@
...
-            Address = property.Address,
+            PropertyType = property.PropertyType ?? string.Empty,
+            NumOfBedrooms = property.NumOfBedrooms.GetValueOrDefault(0),
+            CreationTime = property.CreationTime.GetValueOrDefault(DateTime.Now),
+            Rent = property.Rent.GetValueOrDefault(0),
+            Description = property.Description ?? string.Empty,
+            Address = new PropertyAddress
+            {
+                AddressLine1 = property.AddressLine1!,
+                AddressLine2 = property.AddressLine2 ?? string.Empty,
```
kwunkiuma, line 38: Will likely need to remove this if we plan on adding any null checks 🙂

kwunkiuma (changes requested):
Looking good! Let me know if the comments are better addressed in BNH-34 🙂

web/Views/Landlord/Properties.cshtml
MaxTheComputerer: Yeah I think the nullability of most fields changes in BNH-34, so it might be easier to do it there! 🙂

kwunkiuma:
Looks good to me! Feel free to merge, we can add the null checks in BNH-34 🙂

kwunkiuma (approved):

PR #27, opened 2022-07-20
Bnh 5

blancamc00:
image
Edit buttons added to each user who requested to be admin in the admin list view. Pressing the button marks the user as admin, and no longer pending. 

QUESTION: post method in controller changes the admin and pending status of the users. Inside this method, two other GET methods are called to find the user and to find the modified list of admins and requested admins. Is it okay to call a GET method inside a POST method?

### DIFF
--- a/BricksAndHearts.UnitTests/ControllerTests/Admin/AdminControllerTests.cs
+++ b/BricksAndHearts.UnitTests/ControllerTests/Admin/AdminControllerTests.cs
@@ -56,4 +56,29 @@ public class AdminControllerTests : AdminControllerTestsBase
         A.CallTo(() => AdminService.GetLandlordDisplayList("")).MustHaveHappened();
         result!.ViewData.Model.Should().BeOfType<LandlordListModel?>();
     }
+
+    [Fact]
+    public void GetAdminList_ReturnsViewWithAdminListModel()
+    {
+        // Arrange
+        
+        // Act
+        var result = _underTest.GetAdminList().Result as ViewResult;
+
+        // Assert
+        result!.Model.Should().BeOfType<AdminListModel>();
+    }
+
+    [Fact]
+    public void AcceptAdminRequest_Calls_ApproveAdminAccessRequest()
+    {
+        // Arrange
+        var dummyId = 1;
+
+        // Act
+        var result = _underTest.AcceptAdminRequest(dummyId);
+        
+        // Assert
+        A.CallTo(() => _underTestService.ApproveAdminAccessRequest(dummyId)).MustHaveHappened();
+    }
 }
--- a/BricksAndHearts.UnitTests/ControllerTests/Admin/AdminControllerTests.cs
+++ b/BricksAndHearts.UnitTests/ControllerTests/Admin/AdminControllerTests.cs
@@ -1,4 +1,6 @@
+using System.Security.Claims;
 using BricksAndHearts.ViewModels;
+using FakeItEasy;
 using FluentAssertions;
 using Microsoft.AspNetCore.Http;
 using Microsoft.AspNetCore.Mvc;
@@ -12,34 +14,17 @@ public class AdminControllerTests : AdminControllerTestsBase
     public void Index_WhenCalledByAnonymousUser_ReturnsViewWithLoginLink()
     {
         // Arrange
-        _underTest.ControllerContext = new ControllerContext
+        UnderTest.ControllerContext = new ControllerContext
         {
             HttpContext = new DefaultHttpContext { User = _anonUser }
         };
 
         // Act
-        var result = _underTest.Index() as ViewResult;
+        var result = UnderTest.Index() as ViewResult;
 
         // Assert
         result!.ViewData.Model.Should().BeOfType<AdminViewModel>()
             .Which.CurrentUser.Should().BeNull();
-    }
-    
-    [Fact]
-    public async Task GetAdminList_WhenCalledByAdmin_ReturnsViewWithAdminList()
-    {
-        // Arrange 
-        var adminUser = CreateAdminUserInController(_underTest);
-        _underTest.ControllerContext = new ControllerContext
-        {
-            HttpContext = new DefaultHttpContext { User = new ClaimsPrincipal(adminUser) }
-        };
-
-        // Act
-        var result = await _underTest.GetAdminList() as ViewResult;
-
-        // Assert
-        result!.ViewData.Model.Should().BeOfType<AdminListModel>() ;
     }
 
     [Fact]
@@ -61,10 +46,10 @@ public class AdminControllerTests : AdminControllerTestsBase
     public void GetAdminList_ReturnsViewWithAdminListModel()
     {
         // Arrange
-        MakeUserPrincipalInController(CreateAdminUser(), _underTest);
+        MakeUserPrincipalInController(CreateAdminUser(), UnderTest);
         
         // Act
-        var result = _underTest.GetAdminList().Result as ViewResult;
+        var result = UnderTest.GetAdminList().Result as ViewResult;
 
         // Assert
         result!.ViewData.Model.Should().BeOfType<AdminListModel>();
@@ -74,26 +59,26 @@ public class AdminControllerTests : AdminControllerTestsBase
     public void AcceptAdminRequest_WhenCalled_ApprovesAdminAccessRequest()
     {
         // Arrange
-        MakeUserPrincipalInController(CreateAdminUser(), _underTest);
+        MakeUserPrincipalInController(CreateAdminUser(), UnderTest);
         var dummyId = 1;
 
         // Act
-        var result = _underTest.AcceptAdminRequest(dummyId);
+        var result = UnderTest.AcceptAdminRequest(dummyId);
         
         // Assert
-        A.CallTo(() => adminService.ApproveAdminAccessRequest(dummyId)).MustHaveHappened();
+        A.CallTo(() => AdminService.ApproveAdminAccessRequest(dummyId)).MustHaveHappened();
     }
     
     [Fact]
     public void RejectAdminRequest_WhenCalled_RejectsAdminAccessRequest()
     {
         // Arrange
-        MakeUserPrincipalInController(CreateAdminUser(), _underTest);
+        MakeUserPrincipalInController(CreateAdminUser(), UnderTest);
 
         // Act
-        var result = _underTest.RejectAdminRequest(1);
+        var result = UnderTest.RejectAdminRequest(1);
         
         // Assert
-        A.CallTo(() => adminService.RejectAdminAccessRequest(1)).MustHaveHappened();
+        A.CallTo(() => AdminService.RejectAdminAccessRequest(1)).MustHaveHappened();
     }
 }
--- a/BricksAndHearts.UnitTests/ControllerTests/Admin/AdminControllerTests.cs
+++ b/BricksAndHearts.UnitTests/ControllerTests/Admin/AdminControllerTests.cs
@@ -61,24 +61,26 @@ public class AdminControllerTests : AdminControllerTestsBase
     public void GetAdminList_ReturnsViewWithAdminListModel()
     {
         // Arrange
+        CreateAdminUserInController(UnderTest);
         
         // Act
-        var result = _underTest.GetAdminList().Result as ViewResult;
+        var result = UnderTest.GetAdminList().Result as ViewResult;
 
         // Assert
-        result!.Model.Should().BeOfType<AdminListModel>();
+        result!.ViewData.Model.Should().BeOfType<AdminListModel>();
     }
 
     [Fact]
-    public void AcceptAdminRequest_Calls_ApproveAdminAccessRequest()
+    public void AcceptAdminRequest_WhenCalled_ApprovesAdminAccessRequest()
     {
         // Arrange
+        CreateAdminUserInController(UnderTest);
         var dummyId = 1;
 
         // Act
-        var result = _underTest.AcceptAdminRequest(dummyId);
+        var result = UnderTest.AcceptAdminRequest(dummyId);
         
         // Assert
-        A.CallTo(() => _underTestService.ApproveAdminAccessRequest(dummyId)).MustHaveHappened();
+        A.CallTo(() => AdminService.ApproveAdminAccessRequest(dummyId)).MustHaveHappened();
     }
 }
--- a/BricksAndHearts.UnitTests/ControllerTests/Admin/AdminControllerTests.cs
+++ b/BricksAndHearts.UnitTests/ControllerTests/Admin/AdminControllerTests.cs
@@ -12,13 +12,13 @@ public class AdminControllerTests : AdminControllerTestsBase
     public void Index_WhenCalledByAnonymousUser_ReturnsViewWithLoginLink()
     {
         // Arrange
-        UnderTest.ControllerContext = new ControllerContext
+        _underTest.ControllerContext = new ControllerContext
         {
             HttpContext = new DefaultHttpContext { User = _anonUser }
         };
 
         // Act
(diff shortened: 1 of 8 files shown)


web/Views/Admin/AdminList.cshtml
```diff
@@ -21,7 +21,13 @@
             <tr>
                 <td>@pendingAdmin.GoogleEmail</td>
                 <td>@pendingAdmin.GoogleUserName</td>
-                <td>Actions</td>
+                @*<td><input type="button" class="btn btn-outline-success" value="Accept" onclick="location.href='@Url.Action("AcceptAdminRequest", "Admin", new {userToAccept = pendingAdmin})'"/></td>*@
```
emilypatsko, line 24: Delete code if not being used

web/Views/Admin/AdminList.cshtml
```diff
@@ -21,7 +21,13 @@
...
+                <td>
+                    @using (Html.BeginForm("AcceptAdminRequest", "Admin",new {userToAcceptId = pendingAdmin.Id}, FormMethod.Post))
```
emilypatsko, line 26: Keep an eye on formatting (spaces)
```suggestion
                    @using (Html.BeginForm("AcceptAdminRequest", "Admin", new { userToAcceptId = pendingAdmin.Id }, FormMethod.Post))
```

web/Views/Admin/AdminList.cshtml
```diff
@@ -21,7 +21,13 @@
...
+                    {
+                        <button type="submit"  class="btn btn-outline-success">Accept</button>
```
emilypatsko, line 28: Similar here
```suggestion
                        <button type="submit" class="btn btn-outline-success">Accept</button>
```

emilypatsko (changes requested):

BenRamchandani:
> QUESTION: post method in controller changes the admin and pending status of the users. Inside this method, two other GET methods are called to find the user and to find the modified list of admins and requested admins. Is it okay to call a GET method inside a POST method?

Every public method in a controller represents an endpoint. So `public UserDbModel findUserFromId(int userId)` in `AdminController` will mean anyone can request `/admin/findUserById?userId=<>` to get that information.
If you want a helper method to help keep your code clean then either use private controller methods, or methods on another class, like a service. By convention controllers have logic for parsing and validating requests, and creating responses (which is likely to be endpoint-specific), while services have business logic and things like database updates (which you might want to reuse elsewhere).

Public Controller methods should almost always return `IActionResult` and it's unusual to call one from another, though sometimes makes sense if they are almost identical endpoints, like  `/landlord/me/profile` is a specific case of `/landlord/<id>/profile` in LandlordController.

Does that all make sense? We can maybe chat on a call if not.

web/Controllers/AdminController.cs
```diff
@@ -74,12 +74,27 @@ public IActionResult CancelAdminAccessRequest()
...
         adminListModel.CurrentAdmins = adminLists.CurrentAdmins;
         adminListModel.PendingAdmins = adminLists.PendingAdmins;
-        return View(adminListModel);
+        
+        return View("AdminList", adminListModel);
+    }
+
+    [Authorize(Roles="Admin")]
+    [HttpPost]
+    public async Task<ActionResult> AcceptAdminRequest(AdminListModel userToAcceptList, int userToAcceptId)
```
BenRamchandani, line 92: If this is async, the admin service methods you call should be too, then you can await them

web/Controllers/AdminController.cs
```diff
@@ -74,12 +74,27 @@ public IActionResult CancelAdminAccessRequest()
...
+    {
+        var userToAccept = _adminService.FindUserById(userToAcceptId);
+        
+        _adminService.ApproveAdminAccessRequest(userToAccept);
```
BenRamchandani, line 96: Fine either way, but I'd probably combine these two to be honest, just have one function on AdminService that does both.

I slightly dislike leaking database models out of the service layer, since EF will still track their state and then calling `context.SaveChanges()` can do things you don't expect. Anyway that's mostly me being pedantic

BenRamchandani (changes requested):
This looks much better, couple of small comments

BricksAndHearts.UnitTests/ControllerTests/Admin/AdminControllerTests.cs
```diff
@@ -24,4 +29,33 @@ public void Index_WhenCalledByAnonymousUser_ReturnsViewWithLoginLink()
         result!.ViewData.Model.Should().BeOfType<AdminViewModel>()
             .Which.CurrentUser.Should().BeNull();
     }
+
+    [Fact]
+    public void GetAdminList_ReturnsViewWith_AdminListModel()
+    {
+        // Arrange
+        var fakeAdminService = A.Fake<IAdminService>();
```
BenRamchandani, line 37: You could set this up to return fake data e.g. `A.CallTo(() => fakeAdminService.GetAdminLists()).Returns(stuff)` and then check it's correct. Also fine not to, the code is only two lines so don't want to spent ages writing tests for it

BenRamchandani (changes requested):
Couple of tiny things

BricksAndHearts.UnitTests/ControllerTests/Admin/AdminControllerTests.cs
```diff
@@ -24,4 +31,37 @@ public void Index_WhenCalledByAnonymousUser_ReturnsViewWithLoginLink()
...
```
emilypatsko, line 39: There's already a fake admin service in `AdminControllerTestsBase`, which this class inherits from. You should use that rather than creating a new one here

BricksAndHearts.UnitTests/ControllerTests/Admin/AdminControllerTests.cs
```diff
@@ -24,4 +31,37 @@ public void Index_WhenCalledByAnonymousUser_ReturnsViewWithLoginLink()
...
             .Which.CurrentUser.Should().BeNull();
     }
+
+    [Fact]
+    public void GetAdminList_ReturnsViewWith_AdminListModel()
+    {
+        // Arrange
+        var fakeAdminService = A.Fake<IAdminService>();
+        var fakeListOfAdmins = A.Dummy<Task<(List<UserDbModel> CurrentAdmins, List<UserDbModel> PendingAdmins)>>();
+        var fakeAdminController = new AdminController(null!, null!, fakeAdminService);
```
emilypatsko, line 41: Similarly, `AdminControllerTestsBase` already has a fake `AdminController` called `_underTest` (see the test above this one for an example of it being used)

BricksAndHearts.UnitTests/ControllerTests/Admin/AdminControllerTests.cs
```diff
@@ -24,4 +31,37 @@ public void Index_WhenCalledByAnonymousUser_ReturnsViewWithLoginLink()
...
+        A.CallTo(() => fakeAdminService.GetAdminLists()).Returns(fakeListOfAdmins);
+
+        // Act
+        var result = fakeAdminController.GetAdminList().Result as ViewResult;
+        var resultService = fakeAdminService.GetAdminLists();
+
+        // Assert
+        result!.Model.Should().BeOfType<AdminListModel>();
+        resultService.Should().Be(fakeListOfAdmins);
```
emilypatsko, line 50: We don't want to test the service logic in here - this file is for testing the controller logic. Additionally, we have explicitly said that a call to the service will return a specific result, so this doesn't really test anything

BricksAndHearts.UnitTests/ControllerTests/Admin/AdminControllerTests.cs
```diff
@@ -24,4 +31,37 @@ public void Index_WhenCalledByAnonymousUser_ReturnsViewWithLoginLink()
         result!.ViewData.Model.Should().BeOfType<AdminViewModel>()
             .Which.CurrentUser.Should().BeNull();
     }
+
+    [Fact]
+    public void GetAdminList_ReturnsViewWith_AdminListModel()
```
emilypatsko, line 36: Don't really need the second underscore, reads oddly 

```suggestion
public void GetAdminList_ReturnsViewWithAdminListModel()
```

BricksAndHearts.UnitTests/ControllerTests/Admin/AdminControllerTests.cs
```diff
@@ -24,4 +31,37 @@ public void Index_WhenCalledByAnonymousUser_ReturnsViewWithLoginLink()
...
+        result!.Model.Should().BeOfType<AdminListModel>();
+        resultService.Should().Be(fakeListOfAdmins);
+    }
+
+    [Fact]
+    public void AcceptAdminRequest_Calls_ApproveAdminAccessRequest()
+    {
+        // Arrange
+        var fakeAdminService = A.Fake<IAdminService>();
+        var fakeAdminController = new AdminController(null!, null!, fakeAdminService);
```
emilypatsko, line 58: Same comments as above

BricksAndHearts.UnitTests/ControllerTests/Admin/AdminControllerTests.cs
```diff
@@ -24,4 +31,37 @@ public void Index_WhenCalledByAnonymousUser_ReturnsViewWithLoginLink()
...
+        var dummyId = A.Dummy<int>();
```
emilypatsko, line 59: You can just create a regular integer, no need to fake it. Just pick an arbitrary number

BricksAndHearts.UnitTests/ControllerTests/Admin/AdminControllerTests.cs
```diff
@@ -24,4 +31,37 @@ public void Index_WhenCalledByAnonymousUser_ReturnsViewWithLoginLink()
...
+        var result = fakeAdminController.GetAdminList().Result as ViewResult;
+        var resultService = fakeAdminService.GetAdminLists();
+
+        // Assert
+        result!.Model.Should().BeOfType<AdminListModel>();
+        resultService.Should().Be(fakeListOfAdmins);
+    }
+
+    [Fact]
+    public void AcceptAdminRequest_Calls_ApproveAdminAccessRequest()
```
emilypatsko, line 54: I would rename this test something like `AcceptAdminRequest_WhenCalledWithAdminPermissions_ApproveAdminAccessRequest()`.

We should have tests to check this endpoint executes properly when called by an admin, and that it fails/returns an error/etc. when called by someone without the relevant permissions.

BricksAndHearts.UnitTests/ControllerTests/Admin/AdminControllerTestsBase.cs
```diff
@@ -1,8 +1,14 @@
...
 
 namespace BricksAndHearts.UnitTests.ControllerTests.Admin;
 
 public class AdminControllerTestsBase: ControllerTestsBase
 {
     protected readonly AdminController _underTest = new(null!, null!, null!);
+
+    protected AdminListModel CreateTestAdminListModel()
+    {
+        return new AdminListModel(null!, null!);
```
emilypatsko, line 12: I would probably create this with empty lists rather than nulls

web/Services/AdminService.cs
```diff
@@ -51,4 +53,12 @@ public async Task<(List<UserDbModel> CurrentAdmins, List<UserDbModel> PendingAdm
     {
         return (await GetCurrentAdmins(), await GetPendingAdmins());
     }
+
+    public void ApproveAdminAccessRequest(int userId)
+    {
+        var userToAdmin = _dbContext.Users.Single(u => u.Id == userId);
```
emilypatsko, line 59: In general it's preferable to use the async method where possible. Additionally `SingleOrDefault` is safer as it handles the case where the user ID doesn't exist 
```suggestion
var userToMakeAdmin = await _dbContext.Users.SingleOrDefaultAsync(u => u.Id == userId);
```

web/Controllers/AdminController.cs
```diff
@@ -74,10 +74,22 @@ public IActionResult CancelAdminAccessRequest()
...
-        return View(adminListModel);
+        
+        return View("AdminList", adminListModel);
+    }
+
+    [Authorize(Roles="Admin")]
+    [HttpPost]
+    public ActionResult AcceptAdminRequest(int userToAcceptId)
+    {
+        _adminService.ApproveAdminAccessRequest(userToAcceptId);
```
emilypatsko, line 91: Feels like this should be async - see my comments in `AdminService`

emilypatsko (changes requested):

web/Services/AdminService.cs
blancamc00: image
how can I change this?

BricksAndHearts.UnitTests/ControllerTests/ControllerTestsBase.cs
```diff
@@ -33,4 +33,26 @@ protected BricksAndHeartsUser CreateUnregisteredUserInController(Controller _und
...
+
+        var adminUser = new BricksAndHeartsUser(userDbModel, new List<Claim>(), "google");
+        var adminUserPrincipal = new ClaimsPrincipal(adminUser);
+
+        _underTest.ControllerContext = new ControllerContext
+        {
+            HttpContext = new DefaultHttpContext { User = adminUserPrincipal }
+        };
+        return adminUser;
+    }
```
emilypatsko, line 57: I think this whole method can be reduced to
```
var adminUser = CreateAdminUser()
MakeUserPrincipalInController(adminUser, _underTest)
```
as on `main` it looks like these methods already exist in `ControllerTestsBase`. When was the last time you rebased?

web/Controllers/AdminController.cs
```diff
@@ -74,10 +74,27 @@ public IActionResult CancelAdminAccessRequest()
...
     [Authorize(Roles="Admin")]
     [HttpGet]
-    public async Task<IActionResult> AdminList()
+    [Route("AdminList")]
+    public async Task<IActionResult> GetAdminList()
     {
+        /*if (User.IsInRole("Admin") != true)
+        {
+            return StatusCode(403);
+        }*/
```
emilypatsko, line 83: Remove commented out code

web/Services/AdminService.cs
```diff
@@ -51,4 +53,12 @@ public async Task<(List<UserDbModel> CurrentAdmins, List<UserDbModel> PendingAdm
     {
         return (await GetCurrentAdmins(), await GetPendingAdmins());
     }
+
+    public void ApproveAdminAccessRequest(int userId)
+    {
+        var userToAdmin = _dbContext.Users.SingleOrDefault(u => u.Id == userId)!;
```
emilypatsko, line 59: What happens if this is null (the user doesn't exist)?

emilypatsko (changes requested):

BricksAndHearts.UnitTests/ControllerTests/Admin/AdminControllerTests.cs
```diff
@@ -1,4 +1,12 @@
-﻿using BricksAndHearts.ViewModels;
+﻿using System;
+using System.Collections.Generic;
+using System.Security.Claims;
+using System.Threading.Tasks;
+using BricksAndHearts.Controllers;
+using BricksAndHearts.Database;
+using BricksAndHearts.Services;
```
emilypatsko, line 7: I don't think any of these are being used... could you tidy up unused imports? `CTRL + ALT + O` should do it. or you can just manually delete the grayed out lines in Rider

emilypatsko (changes requested):

emilypatsko:
Also once we've got everything resolved (right before we're ready to merge) I'll ask you to rebase to tidy up the commit history a bit.

BenRamchandani (approved):

BenRamchandani (approved):

PR #28, opened 2022-07-20
BNH-14

SleepyUser:
I've updated the appsettingsdevelopment on Keeper with the changes. 
I have a copy saved locally of the original if there's any issues.

We've added telemetry to the application, and it is viewable in the newly created "bricks-and-hearts-telemetry" in Azure Application Insights. We've checked that our requests on our local environment show up in Azure on its live data.

### DIFF
--- a/web/BricksAndHearts.csproj
+++ b/web/BricksAndHearts.csproj
@@ -8,6 +8,7 @@
     </PropertyGroup>
 
     <ItemGroup>
+        <PackageReference Include="Microsoft.ApplicationInsights.AspNetCore" Version="2.20.0" />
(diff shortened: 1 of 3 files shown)


BenRamchandani (approved):
Looks good, before merging configure the connection string for the App Service in Azure too (lmk if you're not sure how to do that).

Once it's merged check we're getting logs from staging too.

SleepyUser:
Hi Ben, are tests within the Testing Project required for this PR? I've been looking into a way to fake a connection and capture the telemetry being sent to Azure so I can write a test for it, but I'm not sure if this is worth looking into further.

I'll have a look into configuring and get back to you if I can't in about 30 mins.

PR #32, opened 2022-07-21
BNH-19

cec200wire:
Displays the Landlord Charter Status on a landlord's profile page.
LandlordProvidedCharterStatus already existed as a field in the Landlord table - initialised as false - so I used this. It should perhaps be renamed to better reflect what it represents ('CharterStatus' or 'CharterApproved'?), but I was unsure if this change was worth creating new database migrations so have currently left the name unchanged.

### DIFF
--- a/web/Views/Landlord/Profile.cshtml
+++ b/web/Views/Landlord/Profile.cshtml
@@ -63,6 +63,13 @@
     </div>
     
     <h4>ChangeAhead Charter</h4>
-    @*TODO: Database does not include this in this branch yet.*@
+    @if (Model.CharterApproved)
+    {
+        <p>Charter status: Approved</p>
+    }
+    else
+    {
+        <p>Charter status: Pending</p>
+    }
     
 </div>
--- a/web/Database/LandlordDbModel.cs
+++ b/web/Database/LandlordDbModel.cs
@@ -15,6 +15,7 @@ public class LandlordDbModel
 
     public string LandlordStatus { get; set; } = null!;
     public bool LandlordProvidedCharterStatus { get; set; } = false;
+    public bool CharterApproved { get; set; } = false;
 
     public virtual UserDbModel? User { get; set; }
     public virtual List<PropertyDbModel>? Properties { get; set; }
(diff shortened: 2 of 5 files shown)


web/Views/Landlord/Profile.cshtml
```diff
@@ -63,6 +63,13 @@
     </div>
     
     <h4>ChangeAhead Charter</h4>
-    @*TODO: Database does not include this in this branch yet.*@
+    @if (Model.LandlordProvidedCharterStatus)
```
BenRamchandani, line 66: I think this should actually use `LandlordStatus` mainly. Worth chatting to Tom + Blanca who added the two columns for what their intention was, but I think it's:

* LandlordStatus - whether the landlord has been approved by an admin
    * Which is the main thing we care about
* LandlordProvidedCharterStatus - whether the landlord said they have already signed the charter
    * Which I think is just intended to be useful for admins to see so they know whether to check their records, or start the approval process

cec200wire: I've just spoken to them:
- 'LandlordStatus' is a separate thing, referring to the type of landlord (they got it from the existing Landlord Registration Form [Form for Landlords to Complete: (Charities/RSL/PSL/LA/Agencies)](https://docs.google.com/forms/d/e/1FAIpQLSesM0lm2YyY95B8F_XtfakZciti6jDnDhlGAgKIWPqAUZfDkw/viewform?vc=0&c=0&w=1&flr=0))
- 'LandlordProvidedCharterStatus' is indeed whether the landlord said they have signed the charter
- a relevant 'CharterStatus' field (contrary to what was suggested on the ticket) doesn't actually yet exist, so I'll go create that now

cec200wire:
Regarding adding unit tests, I don't think there's any to be usefully added to this branch.

web/Views/Landlord/Profile.cshtml
BenRamchandani: That makes sense, thanks. Would you mind adding comments to those fields in the DB model to explain what they are? (can be separate PR if you want to get this merged)

BenRamchandani (approved):

web/Views/Landlord/Profile.cshtml
cec200wire: Comments added (I didn't think merging this was urgent, so I did it on this pull request)

BenRamchandani (approved):

PR #33, opened 2022-07-21
BNH-12 Sending emails

Chestnut-lol:
No chances to front-end.
When a landlord registers, an email will be sent automatically from changeahead email to my email address.

### DIFF
--- a/BricksAndHearts.UnitTests/ControllerTests/Landlord/LandlordControllerTestsBase.cs
+++ b/BricksAndHearts.UnitTests/ControllerTests/Landlord/LandlordControllerTestsBase.cs
@@ -1,6 +1,5 @@
 ﻿using BricksAndHearts.Controllers;
 using BricksAndHearts.Services;
-using BricksAndHearts.ViewModels;
 using FakeItEasy;
 
 namespace BricksAndHearts.UnitTests.ControllerTests.Landlord;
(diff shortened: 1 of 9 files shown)


BenRamchandani:
Nice! Have you seen it working?

Chestnut-lol:
Yes, I have got the emails like this:
image

This is from a older version and I fixed up the title afterwards.

BenRamchandani:
Awesome, this approach looks great then. Can you tidy it up slightly and we'll get it merged:

* Add the SMTP host and port, and the from address as settings rather than hardcoding them
* Add placeholders to appSettings.json
  * I think ideally add values for smtp.zoo.lan, port 25, no auth, and see if we can get that working, that way everyone can send emails in dev with no setup, but there's also no risk of sending real people emails if anyone ever runs dev against real data (since smtp.zoo.lan will only send to Softwire addresses)

Then let's make tickets for follow on work:

* Email all admins when a landlord registers
* Sending email failing should not cause the endpoint to fail (so wrap in try-catch, and maybe send asynchronously so we're not waiting on it)
* Asking the client what email the real site should send from, and setting that up in the Azure config

Chestnut-lol:
The required tickets have been created :-)

BenRamchandani (approved):
Can you update appSettings.json/appSettings.Development.json (in Keeper) and the readme if required so that sending emails works when developing? smtp.zoo.lan is probably easiest for this.

Once that's done we can get it merged

BenRamchandani (approved):

PR #36, opened 2022-07-21
BNH-36 Admin List table styling

TomHanmer29:
Removed duplicate class tag in tables so that the bootstrap styling worked. Also added col tags to define the column headers better.


BenRamchandani (approved):
LGTM

PR #37, opened 2022-07-21
BNH-43 fixed js bug on site launch

TomHanmer29:
set the javascript function for the registration page call to happen onload for registration page rather than when the site is launched, done with onload in a new html body tag around all the main contents of the page

### DIFF
--- a/web/wwwroot/js/site.js
+++ b/web/wwwroot/js/site.js
@@ -1,5 +1,7 @@
 ﻿let currentTab = 0;
-showTab(currentTab);
+if (window.location.pathname == "/landlord/register"){
+    showTab(currentTab);
+}
 
 function showTab(currentTab) {
     let tabList = document.getElementsByClassName("registerTab");


web/Views/Landlord/Register.cshtml
```diff
@@ -8,96 +8,97 @@
...
-                    <label>Title:</label>
-                    @Html.DropDownList("Title", new List<SelectListItem>
-                    {
-                        new SelectListItem { Text = "Mr", Value = "Mr", Selected = true },
-                        new SelectListItem { Text = "Mrs", Value = "Mrs" },
-                        new SelectListItem { Text = "Miss", Value = "Miss" },
-                        new SelectListItem { Text = "Ms", Value = "Ms" },
-                        new SelectListItem { Text = "Dr", Value = "Dr" }
-                    })
+<body onload="showTab(0)">
```
BenRamchandani, line 11: There's already a `<body>` tag in `_Layout.cshtml` that we're inside of here. I believe the `onload` attribute also works on other elements, can you try just adding it to the existing `<div>` or did that not work?

BenRamchandani (changes requested):

web/Views/Landlord/Register.cshtml
TomHanmer29: I will keep trying to hunt around to see if there is a nice fix for a little while

TomHanmer29:
I have included these changes + extra in BNH-17 since the same bug (which is now fixed) was present in the edit page.

PR #38, opened 2022-07-21
BNH-24

cec200wire:
Database interactions moved from LandlordController to LandlordService via creation of GetLandlordFromId method.

### DIFF
--- a/web/Controllers/LandlordController.cs
+++ b/web/Controllers/LandlordController.cs
@@ -3,7 +3,6 @@ using BricksAndHearts.Services;
 using BricksAndHearts.ViewModels;
 using Microsoft.AspNetCore.Authorization;
 using Microsoft.AspNetCore.Mvc;
-using Microsoft.EntityFrameworkCore;
 
 namespace BricksAndHearts.Controllers;
 
@@ -11,7 +10,6 @@ namespace BricksAndHearts.Controllers;
 [Route("/landlord")]
 public class LandlordController : AbstractController
 {
-    private readonly BricksAndHeartsDbContext _dbContext;
     private readonly ILandlordService _landlordService;
     private readonly IPropertyService _propertyService;
     private readonly ILogger<LandlordController> _logger;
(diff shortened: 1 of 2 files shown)


cec200wire:
I attempted (for quite some time) to write unit tests for LandlordController.Profile, but I believe that the interaction with the database makes creating unit tests difficult - and since the entire point of this ticket was moving database interactions, there perhaps aren't any relevant unit tests to be implemented here.
If this is incorrect, please let me know - I have some messy attempts at unit tests saved on a separate local branch in case the dead ends I was hitting were my fault as opposed to genuine impossibilities.

web/Controllers/LandlordController.cs
```diff
@@ -85,13 +82,13 @@ public async Task<ActionResult> Profile([FromRoute] int id)
             return StatusCode(403);
         }
 
-        var landlord = await _dbContext.Landlords.SingleOrDefaultAsync(l => l.Id == id);
+        var landlord = _landlordService.GetLandlordFromId(id);
         if (landlord == null)
```
BenRamchandani, line 86: The `landlord` variable here is a `Task`, it will never be null. You should await it first.

```suggestion
        var landlord = await _landlordService.GetLandlordFromId(id);
        if (landlord == null)
```

web/Services/LandlordService.cs
```diff
@@ -17,6 +17,7 @@ public enum LandlordRegistrationResult
 
     public Task<LandlordRegistrationResult> RegisterLandlordWithUser(LandlordProfileModel createModel, BricksAndHeartsUser user);
     public List<PropertyDbModel> GetListOfProperties(int landlordId);
+    public Task<LandlordDbModel> GetLandlordFromId(int id);
```
BenRamchandani, line 20: This can return null (or rather, awaiting the Task can return null), so include it in the type signature, and I'd change the method name as well
```suggestion
    public Task<LandlordDbModel?> GetLandlordIfExistsFromId(int id);
```

BenRamchandani (changes requested):
some async trickiness

BenRamchandani (approved):
Looks good. Setting up unit testing for database stuff is whole load of work itself, so definitely fine to skip for now.

cec200wire:
I had to rebase before merging, the conflict resolution was very simple so nothing should have broken but - of course - I now can't merge.

emilypatsko (approved):

PR #39, opened 2022-07-22
BNH-17 landlord edit profile

TomHanmer29:
Landlords can now edit their profiles!
image
image
This uses the same formatting and checks as the landlord register page and thus may have the same issues as on ticket BNH-43 so this is worth noting for future


TomHanmer29:
I forgot to add unit tests! Will close and then reopen once this is done

PR #40, opened 2022-07-22
Bnh 20

cec200wire:
List of unapproved landlords is accessible to admins only. There is a link to this page in the 'Admin' dropdown on the nav-bar, which is only visible to admins.
Hover effect has been added to the table of unapproved landlords to make it clear which landlord is being selected, in order to avoid clicking through to the wrong profile page.

### DIFF
--- a/web/ViewModels/AdminListModel.cs
+++ b/web/ViewModels/AdminListModel.cs
@@ -10,6 +10,6 @@ public class AdminListModel
         PendingAdmins = pendingAdmins;
     }
     
-    public List<UserDbModel>? CurrentAdmins { get; set; }
-    public List<UserDbModel>? PendingAdmins { get; set; }
+    public List<UserDbModel> CurrentAdmins { get; set; }
+    public List<UserDbModel> PendingAdmins { get; set; }
 }
--- a/web/ViewModels/AdminListModel.cs
+++ b/web/ViewModels/AdminListModel.cs
@@ -9,7 +9,6 @@ public class AdminListModel
         CurrentAdmins = currentAdmins;
         PendingAdmins = pendingAdmins;
     }
-    
     public List<UserDbModel> CurrentAdmins { get; set; }
     public List<UserDbModel> PendingAdmins { get; set; }
 }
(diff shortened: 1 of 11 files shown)


cec200wire:
image
Screenshot of new LandlordList page

web/ViewModels/LandlordListModel.cs
```diff
@@ -0,0 +1,7 @@
+﻿using BricksAndHearts.Database;
+namespace BricksAndHearts.ViewModels;
+
+public class LandlordListModel
+{
+    public List<LandlordDbModel>? UnapprovedLandlords { get; set; }
```
kwunkiuma, line 6: Does this need to be nullable? If we don't expect this value to ever be null then it would be best to make it non-nullable 🙂

kwunkiuma (changes requested):
Looks good! Some small comments 🙂

web/ViewModels/LandlordListModel.cs
kwunkiuma: That sounds reasonable 😄 are we sure that we get a null and not an empty list in that scenario?

cec200wire: I'm not certain, but if I don't set it as nullable it causes a build warning which stops me from raising a pull request.

web/ViewModels/AdminListModel.cs
```diff
@@ -9,7 +9,6 @@ public AdminListModel(List<UserDbModel> currentAdmins, List<UserDbModel> pending
         CurrentAdmins = currentAdmins;
         PendingAdmins = pendingAdmins;
     }
-
```
kwunkiuma, line 12: We should keep this newline but make sure it doesn't have any trailing spaces on it 🙂

kwunkiuma (changes requested):
Looks good! Just one tiny thing 😄

web/ViewModels/AdminListModel.cs
cec200wire: Oh, that makes more sense - I was a bit confused by that change.

kwunkiuma (approved):
Looks good to me! Feel free to merge 🙂

PR #42, opened 2022-07-22
BNH-16 bug fix: didn't allow users to sign up

TomHanmer29:
Sign up was previously passing a null value for landlord type as it wasn't updated inside the landlord service file. This is now fixed.


kwunkiuma (commented):
Looks good to me! Can you rebase on main to make sure the build job doesn't fail from any warnings? 🙂

kwunkiuma (approved):
Looks good to me! Feel free to merge 🙂

PR #43, opened 2022-07-22
BNH-21 - approve landlords

cec200wire:
When an admin views the profile page of an unapproved landlord, beneath the 'Charter Status' heading there will be a button to 'Approve Landlord'. This does not appear when the current user is not an admin or when the landlord is already approved.
When clicked, this button changes some entries in the appropriate row of the Landlord table in the database: CharterApproved becomes 'true', ApprovalTime sets to the current DateTime and ApprovalAdminId set to the Id of the current user.
[THERE ARE DATABASE MIGRATIONS IN THIS PULL REQUEST]

Profile page view with button (only change is under 'Charter Status' heading
image

### DIFF
--- a/BricksAndHearts.UnitTests/ControllerTests/Admin/AdminControllerTests.cs
+++ b/BricksAndHearts.UnitTests/ControllerTests/Admin/AdminControllerTests.cs
@@ -1,6 +1,4 @@
-﻿using BricksAndHearts.Controllers;
-using BricksAndHearts.Services;
-using BricksAndHearts.ViewModels;
+﻿using BricksAndHearts.ViewModels;
 using FakeItEasy;
 using FluentAssertions;
 using Microsoft.AspNetCore.Http;
@@ -11,9 +9,6 @@ namespace BricksAndHearts.UnitTests.ControllerTests.Admin;
 
 public class AdminControllerTests : AdminControllerTestsBase
 {
-    private static readonly IAdminService adminService = A.Fake<IAdminService>();
-    private readonly AdminController _underTest = new(null!, adminService);
-    
     [Fact]
     public void Index_WhenCalledByAnonymousUser_ReturnsViewWithLoginLink()
     {
--- a/BricksAndHearts.UnitTests/ControllerTests/Admin/AdminControllerTests.cs
+++ b/BricksAndHearts.UnitTests/ControllerTests/Admin/AdminControllerTests.cs
@@ -1,4 +1,6 @@
-﻿using BricksAndHearts.ViewModels;
+﻿using BricksAndHearts.Controllers;
+using BricksAndHearts.Services;
+using BricksAndHearts.ViewModels;
 using FakeItEasy;
 using FluentAssertions;
 using Microsoft.AspNetCore.Http;
@@ -9,6 +11,9 @@ namespace BricksAndHearts.UnitTests.ControllerTests.Admin;
 
 public class AdminControllerTests : AdminControllerTestsBase
 {
+    private static readonly IAdminService adminService = A.Fake<IAdminService>();
+    private readonly AdminController _underTest = new(null!, adminService);
+    
     [Fact]
     public void Index_WhenCalledByAnonymousUser_ReturnsViewWithLoginLink()
     {
--- a/BricksAndHearts.UnitTests/ControllerTests/Admin/AdminControllerTestsBase.cs
+++ b/BricksAndHearts.UnitTests/ControllerTests/Admin/AdminControllerTestsBase.cs
@@ -1,6 +1,17 @@
-﻿namespace BricksAndHearts.UnitTests.ControllerTests.Admin;
+﻿using BricksAndHearts.Controllers;
+using BricksAndHearts.Services;
+using FakeItEasy;
+
+namespace BricksAndHearts.UnitTests.ControllerTests.Admin;
 
 public class AdminControllerTestsBase: ControllerTestsBase
 {
+    protected readonly IAdminService adminService;
+    protected readonly AdminController _underTest;
 
+    protected AdminControllerTestsBase()
+    {
+        adminService = A.Fake<IAdminService>();
+        _underTest = new(null!, adminService);
+    }
 }
--- a/BricksAndHearts.UnitTests/ControllerTests/Admin/AdminControllerTestsBase.cs
+++ b/BricksAndHearts.UnitTests/ControllerTests/Admin/AdminControllerTestsBase.cs
@@ -1,11 +1,6 @@
-﻿using BricksAndHearts.Controllers;
-using BricksAndHearts.Services;
-using FakeItEasy;
-
-namespace BricksAndHearts.UnitTests.ControllerTests.Admin;
+﻿namespace BricksAndHearts.UnitTests.ControllerTests.Admin;
 
 public class AdminControllerTestsBase: ControllerTestsBase
 {
-    public static readonly IAdminService adminService = A.Fake<IAdminService>();
-    protected readonly AdminController _underTest = new(null!, null!, adminService);
+
 }
--- a/BricksAndHearts.UnitTests/ControllerTests/Landlord/LandlordControllerTests.cs
+++ b/BricksAndHearts.UnitTests/ControllerTests/Landlord/LandlordControllerTests.cs
@@ -1,6 +1,4 @@
-﻿using BricksAndHearts.Controllers;
-using BricksAndHearts.Services;
-using BricksAndHearts.ViewModels;
+﻿using BricksAndHearts.ViewModels;
 using FakeItEasy;
 using FluentAssertions;
 using Microsoft.AspNetCore.Mvc;
@@ -10,10 +8,6 @@ namespace BricksAndHearts.UnitTests.ControllerTests.Landlord;
 
 public class LandlordControllerTests : LandlordControllerTestsBase
 {
-    private static readonly IPropertyService propertyService = A.Fake<IPropertyService>();
-    private static readonly ILandlordService landlordService = A.Fake<ILandlordService>();
-    private readonly LandlordController _underTest = new(null!, null!, landlordService, propertyService);
-
     [Fact]
     public void RegisterGet_CalledByUnregisteredUser_ReturnsRegisterViewWithEmail()
     {
--- a/BricksAndHearts.UnitTests/ControllerTests/Landlord/LandlordControllerTests.cs
+++ b/BricksAndHearts.UnitTests/ControllerTests/Landlord/LandlordControllerTests.cs
@@ -1,4 +1,7 @@
-﻿using BricksAndHearts.ViewModels;
+﻿using BricksAndHearts.Controllers;
+using BricksAndHearts.Services;
+using BricksAndHearts.ViewModels;
+using FakeItEasy;
 using FluentAssertions;
 using Microsoft.AspNetCore.Mvc;
 using Xunit;
@@ -7,6 +10,10 @@ namespace BricksAndHearts.UnitTests.ControllerTests.Landlord;
 
 public class LandlordControllerTests : LandlordControllerTestsBase
 {
+    private static readonly IPropertyService propertyService = A.Fake<IPropertyService>();
+    private static readonly ILandlordService landlordService = A.Fake<ILandlordService>();
+    private readonly LandlordController _underTest = new(null!, null!, landlordService, propertyService);
+
     [Fact]
     public void RegisterGet_CalledByUnregisteredUser_ReturnsRegisterViewWithEmail()
     {
@@ -20,5 +27,20 @@ public class LandlordControllerTests : LandlordControllerTestsBase
         // Assert
         result!.Model.Should().BeOfType<LandlordProfileModel>()
             .Which.Email.Should().Be(unregisteredUser.GoogleEmail);
+    }
+
+    [Fact]
+    public async void ApproveCharter_CallsApproveLandlord()
+    {
+        // Arrange 
+        var adminUser = CreateAdminUser();
+        MakeUserPrincipalInController(adminUser, _underTest);
+        var landlord = CreateLandlordUser();
+
+        // Act
+        var result = await _underTest.ApproveCharter(landlord.Id) as ViewResult;
+
+        // Assert
+        A.CallTo(() => landlordService.ApproveLandlord(landlord.Id, adminUser)).MustHaveHappened();
     }
 }
--- a/BricksAndHearts.UnitTests/ControllerTests/Landlord/LandlordControllerTestsBase.cs
+++ b/BricksAndHearts.UnitTests/ControllerTests/Landlord/LandlordControllerTestsBase.cs
@@ -1,4 +1,4 @@
-﻿using BricksAndHearts.Controllers;
+using BricksAndHearts.Controllers;
 using BricksAndHearts.Services;
 using FakeItEasy;
 
--- a/BricksAndHearts.UnitTests/ControllerTests/Landlord/LandlordControllerTestsBase.cs
+++ b/BricksAndHearts.UnitTests/ControllerTests/Landlord/LandlordControllerTestsBase.cs
@@ -6,6 +6,14 @@ namespace BricksAndHearts.UnitTests.ControllerTests.Landlord;
 
 public class LandlordControllerTestsBase : ControllerTestsBase
 {
-    public static readonly IPropertyService propertyService = A.Fake<IPropertyService>();
-    protected readonly LandlordController _underTest = new(null!, null!, new LandlordService(null!), propertyService,null!);
+    protected readonly IPropertyService propertyService;
+    protected readonly ILandlordService landlordService;
+    protected readonly LandlordController _underTest;
+
+    protected LandlordControllerTestsBase()
+    {
+        landlordService = A.Fake<ILandlordService>();
+        propertyService = A.Fake<IPropertyService>();
+        _underTest = new(null!, landlordService, propertyService, null!);
+    }
 }
--- a/web/Controllers/AdminController.cs
+++ b/web/Controllers/AdminController.cs
@@ -22,7 +22,6 @@ public class AdminController : AbstractController
     {
         _logger = logger;
         _adminService = adminService;
-        _dbContext = dbContext;
     }
 
     public IActionResult Index()
--- a/web/Controllers/AdminController.cs
+++ b/web/Controllers/AdminController.cs
@@ -14,11 +14,9 @@ namespace BricksAndHearts.Controllers;
 public class AdminController : AbstractController
 {
     private readonly IAdminService _adminService;
-    private readonly BricksAndHeartsDbContext _dbContext;
     private readonly ILogger<AdminController> _logger;
 
-    public AdminController(ILogger<AdminController> logger, BricksAndHeartsDbContext dbContext,
-        IAdminService adminService)
+    public AdminController(ILogger<AdminController> logger, IAdminService adminService)
     {
         _logger = logger;
         _adminService = adminService;
--- a/web/Controllers/LandlordController.cs
+++ b/web/Controllers/LandlordController.cs
@@ -15,8 +15,7 @@ public class LandlordController : AbstractController
     private readonly IMailService _mailService;
     private readonly ILogger<LandlordController> _logger;
 
-    public LandlordController(ILogger<LandlordController> logger, BricksAndHeartsDbContext dbContext,
-        ILandlordService landlordService, IPropertyService propertyService, IMailService mailService)
+    public LandlordController(ILogger<LandlordController> logger, ILandlordService landlordService, IPropertyService propertyService, IMailService mailService)
     {
         _logger = logger;
         _landlordService = landlordService;
--- a/web/Controllers/LandlordController.cs
+++ b/web/Controllers/LandlordController.cs
@@ -118,6 +118,15 @@ public class LandlordController : AbstractController
 
         return await Profile(landlordId.Value);
     }
+    
+    [Authorize(Roles="Admin")]
+    [HttpPost]
+    public async Task<ActionResult> ApproveCharter(int landlordId)
+    {
+        var user = GetCurrentUser();
+        await _landlordService.ApproveLandlord(landlordId, user);
+        return RedirectToAction("Profile", "Landlord", new { Id = landlordId });
+    }
 
     [HttpGet]
     [Route("/properties")]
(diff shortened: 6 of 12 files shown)


BricksAndHearts.UnitTests/ControllerTests/ControllerTestsBase.cs
```diff
@@ -25,12 +32,44 @@ protected BricksAndHeartsUser CreateUnregisteredUserInController(Controller _und
...
+    
+    protected BricksAndHeartsUser CreateAdminUser()
+    {
+        var userDbModel = new UserDbModel()
+        {
+            Id = 1,
+            GoogleUserName = "John Doe",
+            GoogleEmail = "test.email@gmail.com",
+            IsAdmin = true,
+            LandlordId = null,
```
kwunkiuma, line 46: Some trailing commas in this file 🙂

BricksAndHearts.UnitTests/ControllerTests/Admin/AdminControllerTestsBase.cs
```diff
@@ -1,8 +1,11 @@
 ﻿using BricksAndHearts.Controllers;
+using BricksAndHearts.Services;
+using FakeItEasy;
 
 namespace BricksAndHearts.UnitTests.ControllerTests.Admin;
 
 public class AdminControllerTestsBase: ControllerTestsBase
 {
-    protected readonly AdminController _underTest = new(null!, null!, null!);
+    public static readonly IAdminService adminService = A.Fake<IAdminService>();
```
kwunkiuma, line 9: I think this should be `protected readonly`, let me know if this doesn't work or if you would like me to clarify the reasoning 🙂

web/Views/Landlord/Profile.cshtml
```diff
@@ -70,6 +70,14 @@
...
+        {
+            @using (Html.BeginForm("ApproveCharter", "Landlord", FormMethod.Post))
+            {
+                <input type="hidden" id="landlordId" name="landlordId" value=@Model.Id>
+                <button type="submit" class="btn custom-btn-primary">Approve Landlord</button>
+            }
+        }
     }
-    
+
```
kwunkiuma, line 82: Extra newline here 🙂

BricksAndHearts.UnitTests/ControllerTests/Landlord/LandlordControllerTestsBase.cs
```diff
@@ -1,9 +1,11 @@
 ﻿using BricksAndHearts.Controllers;
 using BricksAndHearts.Services;
+using FakeItEasy;
 
 namespace BricksAndHearts.UnitTests.ControllerTests.Landlord;
 
 public class LandlordControllerTestsBase : ControllerTestsBase
 {
-    protected readonly LandlordController _underTest = new(null!, null!, new LandlordService(null!));
+    public static readonly ILandlordService landlordService = A.Fake<ILandlordService>();
```
kwunkiuma, line 9: Same comment as the one in `AdminControllerTestsBase` 🙂

kwunkiuma (commented):
Looking good! I've not requested changes so that Ben/Emily can take over looking at this when they're back, but there are some comments you can look at addressing already 🙂

BricksAndHearts.UnitTests/ControllerTests/ControllerTestsBase.cs
cec200wire: They cleared up when I rebased (I removed the trailing commas in the branch I merged yesterday, and made sure to take that version if the file when I rebased this branch).

BricksAndHearts.UnitTests/ControllerTests/Admin/AdminControllerTestsBase.cs
cec200wire: I can make it protected rather than public, but if it isn't declared as static then it can't be accessed in the static context of the next line (creating _underTest, an instance of the AdminController). There might be a way around this, but I'm not sure how complicated it would be - I tried adapting the set-up from the start of the normal AdminController and there wasn't an immediately obvious solution - and also not sure whether the static-ness was part of the problem you were commenting on or not. So for now I've changed it to 'protected static readonly'.

BenRamchandani: Having a `static` fake admin service could lead to problems as it will be reused for all the tests.

Can you move the setup to the constructor instead, creating a new fake admin service and `_underTest` controller in there. It should get run before each test.

BenRamchandani: (and same for landlord tests)

cec200wire: If it's being run before each test, is it therefore ok for the fake admin service to be static (as in, it would be a static service with a fresh copy being made each time it was needed)?

web/Controllers/AdminController.cs
```diff
@@ -89,4 +88,10 @@ public async Task<IActionResult> LandlordList()
         landlordListModel.UnapprovedLandlords = await _adminService.GetUnapprovedLandlords();
         return View(landlordListModel);
     }
+    
+    [HttpPost]
+    public ActionResult ApproveCharter()
+    {
+        return RedirectToAction("Profile", "Landlord", new { Id = 1 });
+    }
```
BenRamchandani, line 96: Can get rid of this I think

web/Services/LandlordService.cs
```diff
@@ -77,4 +78,19 @@ await using (var transaction = await _dbContext.Database.BeginTransactionAsync(I
...
+        landlord.ApprovalTime = DateTime.Now;
+        landlord.ApprovalAdminId = user.Id;
+        await _dbContext.SaveChangesAsync();
+    }
+    
+    public List<PropertyDbModel> GetListOfProperties(int landlordId)
+    {
+        return _dbContext.Properties
+            .Where(p => p.LandlordId == landlordId).ToList();
+    }
```
BenRamchandani, line 95: Is this intentional? Possibly from a merge

web/Services/LandlordService.cs
```diff
@@ -77,4 +78,19 @@ await using (var transaction = await _dbContext.Database.BeginTransactionAsync(I
 
         return ILandlordService.LandlordRegistrationResult.Success;
     }
+
+    public async Task ApproveLandlord(int landlordId,  BricksAndHeartsUser user)
+    {
+        var landlord = _dbContext.Landlords.Single(l =>l.Id == landlordId);
+        landlord.CharterApproved = true;
```
BenRamchandani, line 85: Should we check whether they have been approved already?

I think we probably should. Just throwing an exception in that case is fine I think, though if you prefer you could handle it more nicely with a redirect and flash message.

web/Views/Landlord/Profile.cshtml
```diff
@@ -70,6 +70,13 @@
     else
     {
         <p>Charter status: Pending</p>
+        @if (Model.CurrentUserIsAdmin)
+        {
+            @using (Html.BeginForm("ApproveCharter", "Landlord", FormMethod.Post))
+            {
+                <input type="hidden" id="landlordId" name="landlordId" value=@Model.Id>
+                <button type="submit" class="btn custom-btn-primary">Approve Landlord</button>
```
BenRamchandani, line 78: Is `custom-btn-primary` defined anywhere?

BenRamchandani (changes requested):

web/Services/LandlordService.cs
cec200wire: I just asked Max, apparently it was replaced by a method in PropertyServices - the merge must have accidentally kept both versions.

cec200wire: I've deleted it now, at least from this branch.

web/Views/Landlord/Profile.cshtml
cec200wire: Ah no, I was wondering why that was working weirdly - it's because I looked at some code from bootcamp to figure out how something worked and forgot to change that.

web/Services/LandlordService.cs
cec200wire: The method should only be callable on landlords who have not been approved (the 'Approve' button does not appear on the profile of those who aren't unapproved), but I can add a second check here if you think it's important.

BenRamchandani: I think it's worth having, e.g. someone could have two tabs open, and do it in one and then try and do it again in another, which would then overwrite the approval time. Not the end of the world, but not really desired behaviour.

BenRamchandani (approved):

BricksAndHearts.UnitTests/ControllerTests/Admin/AdminControllerTests.cs
```diff
@@ -9,6 +11,9 @@ namespace BricksAndHearts.UnitTests.ControllerTests.Admin;
 
 public class AdminControllerTests : AdminControllerTestsBase
 {
+    private static readonly IAdminService adminService = A.Fake<IAdminService>();
+    private readonly AdminController _underTest = new(null!, null!, adminService);
```
emilypatsko: Why have these been moved out of `AdminControllerTestsBase`?

BricksAndHearts.UnitTests/ControllerTests/Landlord/LandlordControllerTests.cs
```diff
@@ -7,6 +10,10 @@ namespace BricksAndHearts.UnitTests.ControllerTests.Landlord;
 
 public class LandlordControllerTests : LandlordControllerTestsBase
 {
+    private static readonly IPropertyService propertyService = A.Fake<IPropertyService>();
+    private static readonly ILandlordService landlordService = A.Fake<ILandlordService>();
+    private readonly LandlordController _underTest = new(null!, null!, landlordService, propertyService);
```
emilypatsko, line 15: Feels like these should live in `LandlordControllerTestsBase`

BricksAndHearts.UnitTests/ControllerTests/Landlord/LandlordControllerTests.cs
```diff
@@ -21,4 +28,19 @@ public void RegisterGet_CalledByUnregisteredUser_ReturnsRegisterViewWithEmail()
         result!.Model.Should().BeOfType<LandlordProfileModel>()
             .Which.Email.Should().Be(unregisteredUser.GoogleEmail);
     }
+
+    [Fact]
+    public async void ApproveCharter_CallsApproveLandlord()
```
emilypatsko, line 33:
```suggestion
    public async void ApproveCharter_CalledWithAdminPermissions_CallsApproveLandlord()
```

Don't forget to test this endpoint without admin permissions

BricksAndHearts.UnitTests/ControllerTests/Landlord/LandlordControllerTests.cs
```diff
@@ -21,4 +28,19 @@ public void RegisterGet_CalledByUnregisteredUser_ReturnsRegisterViewWithEmail()
...
+        // Arrange 
+        var adminUser = CreateAdminUser();
+        MakeUserPrincipalInController(adminUser, _underTest);
+        var landlord = CreateLandlordUser();
+
+        // Act
+        var result = await _underTest.ApproveCharter(landlord.Id) as ViewResult;
+
+        // Assert
+        A.CallTo(() => landlordService.ApproveLandlord(landlord.Id, adminUser)).MustHaveHappened();
```
emilypatsko, line 44: Do you want to test anything with the `result`?

BricksAndHearts.UnitTests/ControllerTests/Landlord/LandlordControllerTestsBase.cs
```diff
@@ -1,12 +1,28 @@
...
 using BricksAndHearts.ViewModels;
-using FakeItEasy;
 
 namespace BricksAndHearts.UnitTests.ControllerTests.Landlord;
 
 public class LandlordControllerTestsBase : ControllerTestsBase
 {
-    public static readonly IPropertyService propertyService = A.Fake<IPropertyService>();
-    protected readonly LandlordController _underTest = new(null!, null!, new LandlordService(null!), propertyService);
+    protected PropertyViewModel CreateExamplePropertyViewModel()
```
emilypatsko, line 8: Think this might conflict with Max's work on BNH-34

web/Services/LandlordService.cs
```diff
@@ -85,9 +84,19 @@ await using (var transaction = await _dbContext.Database.BeginTransactionAsync(I
         return _dbContext.Landlords.SingleOrDefaultAsync(l => l.Id == id);
     }
 
-    public List<PropertyDbModel> GetListOfProperties(int landlordId)
+    public async Task ApproveLandlord(int landlordId,  BricksAndHeartsUser user)
     {
-        return _dbContext.Properties
-            .Where(p => p.LandlordId == landlordId).ToList();
+        var landlord = _dbContext.Landlords.Single(l =>l.Id == landlordId);
```
emilypatsko, line 89: Generally preferable to use `SingleOrDefaultAsync` - async preferred, and the 'or default' bit makes sure it handles it if the id doesn't exist
```suggestion
        var landlord = await _dbContext.Landlords.SingleOrDefaultAsync(l => l.Id == landlordId);
```

web/Services/LandlordService.cs
```diff
@@ -85,9 +84,19 @@ await using (var transaction = await _dbContext.Database.BeginTransactionAsync(I
...
+        {
+            landlord.CharterApproved = true;
+            landlord.ApprovalTime = DateTime.Now;
+            landlord.ApprovalAdminId = user.Id;
+            await _dbContext.SaveChangesAsync();
+        }
+        else
+        {
+            throw new Exception("Landlord already approved");
+        }
```
emilypatsko, line 100: Nothing wrong with the logic, but I prefer to shape methods as:

- get info
- if something isn't valid, throw an error
- otherwise continue and do the thing I want to do

This reduces the amount of code blocks. i.e. 

- get landlord
- if already approved, throw error (basically an early return)
- otherwise carry on and approve

web/ViewModels/LandlordProfileModel.cs
```diff
@@ -1,11 +1,14 @@
 ﻿using System.ComponentModel;
 using System.ComponentModel.DataAnnotations;
+using BricksAndHearts.Auth;
 using BricksAndHearts.Database;
 
 namespace BricksAndHearts.ViewModels;
 
 public class LandlordProfileModel
 {
+    public int Id { get; set; }
```
emilypatsko, line 10: Might call this `LandlordId` to differentiate it more clearly from `CurrentUserId` below

emilypatsko (changes requested):

BricksAndHearts.UnitTests/ControllerTests/Admin/AdminControllerTests.cs
cec200wire: There was a potential issue with a static fake admin service being reused for every test - Ben suggested moving the setup to the constructor instead (see the conversation under Frank's suggestions above, I've also screenshotted it and put it here for convenience). I'm not actually certain that I've now put them in the intended place - I think maybe they should be in each individual test? - but that's why they've moved out of AdminControllerTestsBase.
image

BricksAndHearts.UnitTests/ControllerTests/Landlord/LandlordControllerTests.cs
cec200wire: I looked into unit testing things which should require authorisation, and from what I could find it's not actually possible - the unit test calls the method and ignores any [Authorise] labels above it. The only way of checking authorisation is to write a test that literally just checks that any relevant method/endpoint for the [Authorise] label, and I'm not sure if this is worth doing - if it is, it's probably worth raising a separate ticket as it would be a shared test, listing anything which requires authorisation, rather than specific to this endpoint.

BricksAndHearts.UnitTests/ControllerTests/Landlord/LandlordControllerTestsBase.cs
cec200wire: I rebased very recently (right before I left yesterday afternoon) so it should be fine.

BricksAndHearts.UnitTests/ControllerTests/Admin/AdminControllerTests.cs
emilypatsko: I think he meant to put in the constructor of `AdminControllerTestsBase`, like:
```
public class AdminControllerTestsBase: ControllerTestsBase
{
    protected readonly IAdminService AdminService;
    protected readonly AdminController UnderTest;

    protected AdminControllerTestsBase()
    {
        AdminService = A.Fake<IAdminService>();
        UnderTest = new(null!, null!, AdminService);
    }
}
```
otherwise it doesn't seem the tests base is really serving any purpose. @BenRamchandani have I understood what you were suggesting?

emilypatsko: (and same for any other controller tests base files)

BricksAndHearts.UnitTests/ControllerTests/Landlord/LandlordControllerTests.cs
emilypatsko: See my above comments on `AdminControllerTests` :)

BricksAndHearts.UnitTests/ControllerTests/Landlord/LandlordControllerTests.cs
cec200wire: For example, I tried adding the line 
result!.ViewData.Model.Should().BeOfType<LandlordProfileModel?>();
but result comes back as null, so the test then fails.
I guess since result isn't currently used I could get rid of it entirely, but I think that once database-related unit tests are working it would be good to go back and add an appropriate test for the result here - leaving the unused variable here might remind whoever ends up adding missing unit tests that this unit test could do with updating.

web/Services/LandlordService.cs
```diff
@@ -85,9 +84,16 @@ await using (var transaction = await _dbContext.Database.BeginTransactionAsync(I
         return _dbContext.Landlords.SingleOrDefaultAsync(l => l.Id == id);
     }
 
-    public List<PropertyDbModel> GetListOfProperties(int landlordId)
+    public async Task ApproveLandlord(int landlordId,  BricksAndHeartsUser user)
     {
-        return _dbContext.Properties
-            .Where(p => p.LandlordId == landlordId).ToList();
+        var landlord = await GetLandlordIfExistsFromId(landlordId);
```
emilypatsko, line 89: What if the landlord doesn't exist?

emilypatsko (dismissed):

web/ViewModels/LandlordProfileModel.cs
```diff
@@ -50,7 +57,9 @@ public static LandlordProfileModel FromDbModel(LandlordDbModel landlord)
             Title = landlord.Title,
             LandlordStatus = landlord.LandlordStatus,
             LandlordProvidedCharterStatus = landlord.LandlordProvidedCharterStatus,
-            CharterApproved = landlord.CharterApproved
+            CharterApproved = landlord.CharterApproved,
+            CurrentUserIsAdmin = user.IsAdmin,
+            CurrentUserId = user.Id
```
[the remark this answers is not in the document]
cec200wire: CurrentUserIsAdmin will slightly affect the displayed profile (things like the 'Approve Landlord' button will only appear if the current user is an admin). As far as I know, incorporating this property into the LandlordProfileModel is the easiest (if not only) way send this information to the Profile page and thus adjust the view accordingly.

CurrentUserId *was* included so that, if the user is an admin and they approve the landlord, the id of the admin who performs this action can be saved to the database (as specified in the ticket). However, having just re-checked the code, this doesn't actually seem to be needed - I think when I first made changes to the Profile Model, I forgot that it was possible to GetCurrentUser() between pressing the 'Approve Landlord' button and the database changes occurring. I've deleted this property now.

web/Services/LandlordService.cs
```diff
@@ -85,9 +84,16 @@ await using (var transaction = await _dbContext.Database.BeginTransactionAsync(I
...
-    public List<PropertyDbModel> GetListOfProperties(int landlordId)
+    public async Task ApproveLandlord(int landlordId,  BricksAndHeartsUser user)
     {
-        return _dbContext.Properties
-            .Where(p => p.LandlordId == landlordId).ToList();
+        var landlord = await GetLandlordIfExistsFromId(landlordId);
+        if (landlord!.CharterApproved)
+        {
+            throw new Exception("Landlord already approved");
+        }
```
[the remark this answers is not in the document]
cec200wire: The result does actually change slightly - it would update the ApprovalTime, and possibly also the ApprovalAdminId, if a landlord was re-approved. Ben pointed out this slight difference and suggested it would be worth just adding this quick check, which made sense to me.

web/Controllers/LandlordController.cs
```diff
@@ -106,6 +106,15 @@ public async Task<ActionResult> MyProfile()
...
     }
+    
+    [Authorize(Roles="Admin")]
+    [HttpPost]
+    public async Task<ActionResult> ApproveCharter(int landlordId)
+    {
+        var user = GetCurrentUser();
+        await _landlordService.ApproveLandlord(landlordId, user);
+        return RedirectToAction("Profile", "Landlord", new { Id = landlordId });
+    }
```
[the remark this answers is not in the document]
BenRamchandani: I don't feel strongly about it.. I'd probably leave it in Landlord controller myself, but being consistent is the only thing that really matters (so actions that admins take on landlords should either all go in admin or all go in landlord)

web/Services/LandlordService.cs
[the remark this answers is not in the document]
BenRamchandani: @emilypatsko you've been making too many APIs, you can't `PUT` from an HTML form 😛

web/ViewModels/LandlordProfileModel.cs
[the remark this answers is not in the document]
BenRamchandani: In this specific case you could do `@if (User.IsInRole("Admin"))` in the Razor view.

Probably the "right" way to do things like this in general where you're adding extra page-specific data would be to make a new "LandlordProfilePageViewModel" that includes this model as a property along with "CurrentUserIsAdmin".

If we end up adding several extra properties here then it's worth changing, but otherwise I wouldn't worry

BenRamchandani (changes requested):
One more thing!

BenRamchandani (approved):

PR #45, opened 2022-07-22
BNH-17 Edit landlord profile

TomHanmer29:
Allowed users to edit the landlord profile and made unit tests to go with it. The interface is reused from landlord sign up so likely has the same JS error on the loading of the website, this should be addressed in the same/similar ticket to BNH-43
image
image


web/Services/LandlordService.cs
```diff
@@ -79,4 +83,26 @@ public List<PropertyDbModel> GetListOfProperties(int landlordId)
...
+
+    public LandlordDbModel UpdateEditedLandlord(LandlordDbModel createModel)
+    {
+        var editedLandlord = _dbContext.Landlords.SingleOrDefault(l => l.Id == createModel.Id);
+        editedLandlord!.Title = createModel.Title;
+        editedLandlord.FirstName = createModel.FirstName;
+        editedLandlord.LastName = createModel.LastName;
+        editedLandlord.CompanyName = createModel.CompanyName;
+        /*TODO: check that the new email input is not currently registered*/
+        editedLandlord.Email = createModel.Email;
```
kwunkiuma, line 95: Some trailing spaces on this line 🙂

web/wwwroot/css/site.css
```diff
@@ -58,6 +58,28 @@ input.invalid {
   background-color: #04AA6D;
 }
 
-.tables{
+/*.tables{
```
kwunkiuma, line 61: What's the reason that this is commented out? 🙂

kwunkiuma (commented):
Looks good! Some small comments for now 🙂

web/wwwroot/css/site.css
TomHanmer29: I think this is leftover from a different branch :)

web/Services/LandlordService.cs
```diff
@@ -77,4 +81,41 @@ await using (var transaction = await _dbContext.Database.BeginTransactionAsync(I
 
         return ILandlordService.LandlordRegistrationResult.Success;
     }
+
+    public List<PropertyDbModel> GetListOfProperties(int landlordId)
+    {
+        return _dbContext.Properties
+            .Where(p => p.LandlordId == landlordId).ToList();
+    }
```
BenRamchandani, line 89: Is this supposed to be here?

BenRamchandani (changes requested):

web/Controllers/LandlordController.cs
```diff
@@ -153,4 +153,20 @@ public ActionResult AddNewProperty([FromForm] PropertyViewModel newPropertyModel
 
         return RedirectToAction("ViewProperties");
     }
+
+    [HttpGet]
+    [Route("edit")]
+    public ActionResult EditProfilePage(string userEmail)
+    {
+        var landlord = _landlordService.GetLandlordFromEmail(userEmail);
```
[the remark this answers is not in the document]
TomHanmer29: Yeah this is a very good point - I'll refactor the whole thing to only work on Id's :) I think cause I took this ticket over from Blanca half way through I just carried on with the methods she'd put in in the first place.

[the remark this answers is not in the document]
TomHanmer29: This should mean I can reuse the whole signup form more easily too hopefully, since i'll be putting in the correct type of model this time

TomHanmer29:
This ticket also now addresses BNH-43 since the same error was present in the edit page and thought I'd fix it here rather than opening another tiny ticket. I don't know how to delete a ticket from the JIRA board, but ticket 43 should probably be removed if this gets approved?

web/Views/Shared/_RegisterPagePartial.cshtml
```diff
@@ -0,0 +1,82 @@
...
+<div class="registerTab" id="Terms and conditions">
+    @*tickbox and submit button*@
+</div>
+
+<hr/>
+<button type="button" class="btn btn-outline-primary" id="prevBtn" onclick="nextPrev(-1)">Previous</button>
+<button type="button" class="btn btn-outline-primary" id="nextBtn" onclick="nextPrev(1)">Next</button>
+
+<!-- Circles which indicates the steps of the form: -->
+<div style="text-align:center;margin-top:40px;">
```
emilypatsko, line 76: Are there bootstrap classes which could do this instead of inline styles?

web/wwwroot/css/site.css
```diff
@@ -58,6 +58,29 @@ input.invalid {
...
+  float: right;
+  text-align: right;
+  margin-left: 80%;
+}
+
+.bg-primary p-2 rounded-1 {
+  position: fixed;
+  right: 70%;
+}
+
```
emilypatsko, line 83: what are these classes for? Just out of curiosity. I'm hesitant to do too much with CSS as we're still waiting on designs, and a lot of stuff can be done with Bootstrap.

emilypatsko (changes requested):

web/Services/LandlordService.cs
```diff
@@ -79,15 +80,33 @@ await using (var transaction = await _dbContext.Database.BeginTransactionAsync(I
...
+        _dbContext.SaveChanges();
+    }
 
-    public List<PropertyDbModel> GetListOfProperties(int landlordId)
+    public bool CheckForDuplicateEmail(LandlordProfileModel createModel, int? selectedUserId)
     {
-        return _dbContext.Properties
-            .Where(p => p.LandlordId == landlordId).ToList();
+        var editedLandlord = _dbContext.Landlords.Single(l => l.Id == selectedUserId);
+        if (_dbContext.Landlords.SingleOrDefault(l => l.Email == createModel.Email)==null || editedLandlord.Email==createModel.Email)
```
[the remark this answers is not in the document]
TomHanmer29: I did know about this but I thought the if statement might make the code easier to read. Is it generally better practice if you're just returning a bool from an if statement to just return the statement?

web/Controllers/LandlordController.cs
```diff
@@ -133,4 +132,29 @@ public IActionResult ViewProperties()
...
+    public ActionResult EditProfileUpdate([FromForm] LandlordProfileModel createModel)
+    {
+        var user = GetCurrentUser();
+        var isEmailDuplicated = _landlordService.CheckForDuplicateEmail(createModel, user.LandlordId);
+        if (isEmailDuplicated)
+        {
+            _logger.LogWarning("Email already registered {Email}", createModel.Email);
+            ModelState.AddModelError("Email", "Email already registered");
+            return View("EditProfilePage", createModel);
+        }
```
BenRamchandani, line 154: I think rather than getting the landlord id from the user, you should supply it in the route or form values, then check the user has permission to edit that landlord. I think that would be slightly clearer to understand and has a couple of advantages:

* Admins can then use the same route to edit landlord's profiles
* If we ever want multiple users per landlord it would just work

TomHanmer29: This was a change I was working on in BNH-45 (Admins edit landlord profile) :) It's why I asked about adding landlord id to the landlord profile model! I can add it in to this ticket if you think that would be a good idea or I can change it in ticket 45, depends what you think.

web/wwwroot/css/site.css
TomHanmer29: I think Blanca added these for the buttons on the edit page before I took over this branch. I will redo the buttons using bootstrap! :)

web/Controllers/LandlordController.cs
TomHanmer29: I think after thinking abt it it makes more sense to include it here, since ticket 45 just says "admin edit functionality may already be implemented so could just need testing" and that way I dont have to wait for another pull request to be approved.

web/Services/LandlordService.cs
```diff
@@ -100,4 +109,26 @@ public async Task ApproveLandlord(int landlordId,  BricksAndHeartsUser user)
...
+        await _dbContext.SaveChangesAsync();
+        return ILandlordService.LandlordRegistrationResult.Success;
+    }
+
+    public bool CheckForDuplicateEmail(LandlordProfileModel createModel, int? selectedUserId)
+    {
+        var editedLandlord = _dbContext.Landlords.Single(l => l.Id == selectedUserId);
+        return _dbContext.Landlords.SingleOrDefault(l => l.Email == createModel.Email) != null 
+               && editedLandlord.Email != createModel.Email;
+    }
```
BenRamchandani, line 133: There's some odd names here that make this a bit confusing.

* `createModel` isn't really a create model here, it's an `editModel`, I guess?
* `selectedUserId` is a landlord id, not a user id. I also don't think we need it at all now that `LandlordProfileModel` has the id in itself
* The name `CheckForDuplicateEmail` doesn't tell us what the return value means, I think it would be clearer to name it something like `EditWouldCreateDuplicateEmail`

web/Services/LandlordService.cs
```diff
@@ -78,20 +81,26 @@ await using (var transaction = await _dbContext.Database.BeginTransactionAsync(I
...
     public Task<LandlordDbModel?> GetLandlordIfExistsFromId(int id)
     {
         return _dbContext.Landlords.SingleOrDefaultAsync(l => l.Id == id);
     }
 
+    public List<PropertyDbModel> GetListOfProperties(int landlordId)
+    {
+        return _dbContext.Properties
+            .Where(p => p.LandlordId == landlordId).ToList();
+    }
```
BenRamchandani, line 94: Is this supposed to be here?

web/Controllers/LandlordController.cs
```diff
@@ -141,4 +141,28 @@ public IActionResult ViewProperties()
...
+        var isEmailDuplicated = _landlordService.CheckForDuplicateEmail(createModel, createModel.LandlordId);
+        if (isEmailDuplicated)
+        {
+            _logger.LogWarning("Email already registered {Email}", createModel.Email);
+            ModelState.AddModelError("Email", "Email already registered");
+            return View("EditProfilePage", createModel);
+        }
+
+        await _landlordService.EditLandlordDetails(createModel, createModel.LandlordId);
+        _logger.LogInformation("Successfully created landlord for user {UserId}", createModel.LandlordId);
```
BenRamchandani, line 165: Should be edited, and would be nice to include the landlord id too

web/Controllers/LandlordController.cs
```diff
@@ -141,4 +141,28 @@ public IActionResult ViewProperties()
         var listOfProperties = databaseResult.Select(PropertyViewModel.FromDbModel).ToList();
         return View("Properties", new PropertiesDashboardViewModel(listOfProperties));
     }
+    
+    [HttpGet]
+    [Route("edit")]
+    public ActionResult EditProfilePage(LandlordProfileModel currentProfileModel)
+    {
+        return View("EditProfilePage", currentProfileModel);
+    }
```
BenRamchandani, line 150: This relies on the `LandlordProfileModel` already being populated in the current session. Try restarting the server and then going straight to `/landlord/edit`. We should load the landlord record from the database in this method.

It would also be good to follow the same URL convention that the profile uses, i.e. `/landlord/id:int/profile` and `/landlord/me/profile` - that makes it easy for us to extend this to admins that don't have a specific landlord associated with them. That's a bit more work though, so probably better to leave for BNH-45

web/Controllers/LandlordController.cs
```diff
@@ -141,4 +141,28 @@ public IActionResult ViewProperties()
...
+    public ActionResult EditProfilePage(LandlordProfileModel currentProfileModel)
+    {
+        return View("EditProfilePage", currentProfileModel);
+    }
+
+    [HttpPost]
+    [Route("edit")]
+    public async Task<ActionResult> EditProfileUpdate([FromForm] LandlordProfileModel createModel)
+    {
+        var isEmailDuplicated = _landlordService.CheckForDuplicateEmail(createModel, createModel.LandlordId);
```
BenRamchandani, line 156: We need to check the user has permission to edit this landlord

BenRamchandani (changes requested):

web/Controllers/LandlordController.cs
TomHanmer29: Just a couple of quick clarifications: 
What do you mean by restarting the server? Do you just mean get rid of the current page or is there some kind of method I should use?
Also would you like me to do both of these changes in BNH-45 or just the URL changes in BNH-45
Thanks :)

BenRamchandani: Press stop and start in Rider.

Just the URL changes, should make the change to load the landlord in this method now

web/Services/LandlordService.cs
```diff
@@ -96,4 +99,26 @@ public async Task ApproveLandlord(int landlordId,  BricksAndHeartsUser user)
...
+    {
+        var landlordToEdit = await _dbContext.Landlords.SingleAsync(l => l.Id == editModel.LandlordId);
+        landlordToEdit.Title = editModel.Title;
+        landlordToEdit.FirstName = editModel.FirstName;
+        landlordToEdit.LastName = editModel.LastName;
+        landlordToEdit.CompanyName = editModel.CompanyName;
+        landlordToEdit.Email = editModel.Email;
+        landlordToEdit.Phone = editModel.Phone;
+
+        _dbContext.Update(landlordToEdit);
```
BenRamchandani, line 113: I wouldn't expect this to be necessary, does it not work without it?

My understanding is that it enables Entity Framework's change tracking on an object, but that should be enabled by default for things read from the database unless you do `.AsNoTracking()`

web/wwwroot/js/site.js
```diff
@@ -1,15 +1,26 @@
-﻿let currentTab = 0;
-showTab(currentTab);
+﻿var currentTab = 0;
+
+function setCurrentTab(inputTab){
+    currentTab = inputTab
+    localStorage.setItem("someVarKey", currentTab);
```
BenRamchandani, line 5: Lets rename "someVarKey" while we're here

BenRamchandani (changes requested):
Couple of small things

web/Services/LandlordService.cs
TomHanmer29: Yep! You're right, I guess I just didn't try testing this without this line ever :)

web/Controllers/LandlordController.cs
```diff
@@ -141,4 +142,32 @@ public IActionResult ViewProperties()
...
+        if (landlordFromDb == null) return StatusCode(404);
+        if (user.LandlordId != landlordFromDb.Id && !user.IsAdmin) return StatusCode(403);
+        return View("EditProfilePage", LandlordProfileModel.FromDbModel(landlordFromDb, user));
+    }
+
+    [HttpPost]
+    [Route("edit")]
+    public async Task<ActionResult> EditProfileUpdate([FromForm] LandlordProfileModel editModel)
+    {
+        var isEmailDuplicated = _landlordService.CheckForDuplicateEmail(editModel);
```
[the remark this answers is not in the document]
TomHanmer29: If I'm going to check both this and the other edit method using the same conditions, is it worth putting that check into a method since I'd be using it twice? If so, should that method go in the controller or the service?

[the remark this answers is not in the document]
TomHanmer29: I wrote a comment here but realised I was being silly so have deleted it

web/Controllers/LandlordController.cs
```diff
@@ -141,4 +141,34 @@ public IActionResult ViewProperties()
...
+        var user = GetCurrentUser();
+        var landlordFromDb =  await _landlordService.GetLandlordIfExistsFromId(landlordId);
+        if (landlordFromDb == null) return StatusCode(404);
+        if (user.LandlordId != landlordFromDb.Id && !user.IsAdmin) return StatusCode(403);
+        return View("EditProfilePage", LandlordProfileModel.FromDbModel(landlordFromDb, user));
+    }
+
+    [HttpPost]
+    [Route("edit")]
+    public async Task<ActionResult> EditProfileUpdate([FromForm] LandlordProfileModel editModel)
```
BenRamchandani, line 158: One more thing, should check `ModelState.IsValid` here. Fine to put in your other MR if you want to get this merged, sorry didn't think of it earlier.

BenRamchandani (approved):

web/Controllers/LandlordController.cs
TomHanmer29: Ahhh yeah good point! I will merge and add this to the other MR

PR #46, opened 2022-07-25
BNH-29

SleepyUser:
Extracted methods for reuse within the same class, and then shifted them to the abstract superclass for reuse in future controllers

### DIFF
--- a/web/Controllers/AbstractController.cs
+++ b/web/Controllers/AbstractController.cs
@@ -24,14 +24,6 @@ public abstract class AbstractController : Controller
                 $"Successfully {requestAction}"));
     }
 
-    protected void LoggerAlreadyAdminWarning(ILogger logger, BricksAndHeartsUser user)
-    {
-        FlashMessage(logger,
-            ($"User {user.Id} already an admin",
-                "danger",
-                "Already an admin"));
-    }
-
     protected void FlashMessage(ILogger logger, (string logInfo, string flashtype, string flashmessage) flash)
     {
         logger.LogInformation(flash.logInfo);
--- a/web/Controllers/AbstractController.cs
+++ b/web/Controllers/AbstractController.cs
@@ -11,4 +11,27 @@ public abstract class AbstractController : Controller
 
         return (BricksAndHeartsUser)User.Identity;
     }
+    
+    protected void FlashRequestSuccess(ILogger logger, BricksAndHeartsUser user, string requestType)
+    {
+        FlashMessage(logger,
+            ($"Successfully {requestType} for user {user.Id}",
+                "success",
+                $"Successfully {requestType}"));
+    }
+
+    protected void LoggerAlreadyAdminWarning(ILogger logger, BricksAndHeartsUser user)
+    {
+        FlashMessage(logger,
+            ($"User {user.Id} already an admin", 
+                "danger", 
+                "Already an admin"));
+    }
+
+    protected void FlashMessage(ILogger logger, (string logInfo,string flashtype, string flashmessage) flash)
+    {
+        logger.LogInformation(flash.logInfo);
+        TempData["FlashType"] = flash.flashtype;
+        TempData["FlashMessage"] = flash.flashmessage;
+    }
 }
--- a/web/Controllers/AbstractController.cs
+++ b/web/Controllers/AbstractController.cs
@@ -1,5 +1,9 @@
-﻿using BricksAndHearts.Auth;
+﻿#region
+
+using BricksAndHearts.Auth;
 using Microsoft.AspNetCore.Mvc;
+
+#endregion
 
 namespace BricksAndHearts.Controllers;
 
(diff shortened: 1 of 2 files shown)


web/Controllers/AdminController.cs
```diff
@@ -35,40 +36,33 @@ public IActionResult RequestAdminAccess()
...
-        _logger.LogInformation("Successfully requested admin access for user {UserId}",user.Id);
-        TempData["FlashType"] = "success";
-        TempData["FlashMessage"] = "Successfully requested admin access";
-
+        
+        FlashRequestSuccess(_logger, user, "requested admin access");
+        
         return RedirectToAction(nameof(Index));
     }
+
```
kwunkiuma, line 50: Extra newline here 🙂

web/Controllers/AdminController.cs
```diff
@@ -80,4 +74,5 @@ public async Task<IActionResult> AdminList()
         AdminListModel adminListModel = new AdminListModel(adminLists.CurrentAdmins, adminLists.PendingAdmins);
         return View(adminListModel);
     }
+
```
kwunkiuma, line 77: Unecessary newline here 🙂

web/Controllers/AbstractController.cs
```diff
@@ -11,4 +11,27 @@ protected BricksAndHeartsUser GetCurrentUser()
 
         return (BricksAndHeartsUser)User.Identity;
     }
+    
+    protected void FlashRequestSuccess(ILogger logger, BricksAndHeartsUser user, string requestType)
```
kwunkiuma, line 15: I think our input isn't much of a type, maybe `requestAction` would be a better name? 🙂

web/Controllers/AbstractController.cs
```diff
@@ -11,4 +11,27 @@ protected BricksAndHeartsUser GetCurrentUser()
...
+
+    protected void LoggerAlreadyAdminWarning(ILogger logger, BricksAndHeartsUser user)
+    {
+        FlashMessage(logger,
+            ($"User {user.Id} already an admin", 
+                "danger", 
+                "Already an admin"));
+    }
+
+    protected void FlashMessage(ILogger logger, (string logInfo,string flashtype, string flashmessage) flash)
```
kwunkiuma, line 31: Missing a space after `logInfo` 🙂

kwunkiuma (commented):
Looking good! Haven't requested changes so Ben/Emily can take over but left comments that can be addressed 🙂

web/Controllers/AbstractController.cs
```diff
@@ -11,4 +11,27 @@ protected BricksAndHeartsUser GetCurrentUser()
...
+        FlashMessage(logger,
+            ($"Successfully {requestType} for user {user.Id}",
+                "success",
+                $"Successfully {requestType}"));
+    }
+
+    protected void LoggerAlreadyAdminWarning(ILogger logger, BricksAndHeartsUser user)
+    {
+        FlashMessage(logger,
+            ($"User {user.Id} already an admin",
```
kwunkiuma, line 26: Some trailing spaces on this and the line below 🙂

web/Controllers/AbstractController.cs
```diff
@@ -11,4 +15,27 @@ protected BricksAndHeartsUser GetCurrentUser()
...
+                $"Successfully {requestAction}"));
+    }
+
+    protected void LoggerAlreadyAdminWarning(ILogger logger, BricksAndHeartsUser user)
+    {
+        FlashMessage(logger,
+            ($"User {user.Id} already an admin",
+                "danger",
+                "Already an admin"));
+    }
```
emilypatsko, line 33: Is anything other than the `AdminController` going to use this method? If not I would keep it there. The other two methods make sense to commonise to me though :)

emilypatsko (dismissed):
One small comment, but otherwise this is fine

web/Controllers/AbstractController.cs
SleepyUser: I think you're right, and it can easily be taken out again if it is required though

web/Controllers/AbstractController.cs
```diff
@@ -1,6 +1,10 @@
-﻿using BricksAndHearts.Auth;
+﻿#region
+
+using BricksAndHearts.Auth;
 using Microsoft.AspNetCore.Mvc;
 
+#endregion
```
BenRamchandani, line 6: Is this just to hide the imports? Rider should be able to do that anyway

web/Controllers/AbstractController.cs
```diff
@@ -11,4 +15,19 @@ protected BricksAndHeartsUser GetCurrentUser()
...
+    {
+        FlashMessage(logger,
+            ($"Successfully {requestAction} for user {user.Id}",
+                "success",
+                $"Successfully {requestAction}"));
+    }
+
+    protected void FlashMessage(ILogger logger, (string logInfo, string flashtype, string flashmessage) flash)
+    {
+        logger.LogInformation(flash.logInfo);
```
BenRamchandani, line 29: I think having the log statement here is probably overkill, but doesn't really matter, we can always take it out later if we don't find it useful

BenRamchandani (approved):

BenRamchandani:
Also random thing, can you include a very brief title in PRs, probably just the ticket title, so I don't have to look up the ticket 😛

PR #48, opened 2022-07-25
BNH 33 postcode autocomplete

Chestnut-lol:
This branch is rebased on BNH-34.

It adds the additional functionality of postcode autofill when filling in "Add New Property" form.


BenRamchandani:
This has some changes from another branch on I think, would you mind rebasing on to main?

Chestnut-lol:
Yes I have rebased it onto main, should work pretty well now.

web/Services/ApiService.cs
```diff
@@ -0,0 +1,100 @@
...
+    {
+        _logger = logger;
+        _options = options;
+    }
+
+    public async Task<string> MakeApiRequestToAzureMaps(string postalCode)
+    {
+        string uri =
+            $"https://atlas.microsoft.com/search/address/structured/json?api-version=1.0&countryCode=GB&postalCode={postalCode}&subscription-key={_options.Value.SubscriptionKey}";
+        string responseBody = String.Empty;
```
BenRamchandani, line 31: Can you run Rider's inspector over your code?
image
It should come up with a bunch of small things that are usually (though not always) worth fixing, for example using var here, and giving names to the format options in the log statement below, which is important as they come through to Application Insights.

If you're not sure what any of them mean or why you might want to do them let me know.

web/ViewModels/PostcodeApiResponseViewModel.cs
```diff
@@ -0,0 +1,17 @@
+﻿using Newtonsoft.Json;
+
+namespace BricksAndHearts.ViewModels;
+
+public class PostcodeApiResponseViewModel
```
BenRamchandani, line 5: This isn't a view model (we're not using it in a view). I think move it into the services folder, or even inside ApiService.

web/Services/ApiService.cs
```diff
@@ -0,0 +1,100 @@
...
+            return model;
+        }
+
+        BricksAndHearts.ViewModels.Results results = postcodeResponse.ListOfResults[0];
+        if (results.Address == null)
+        {
+            return model;
+        }
+        
+        if (results.Address.Keys.ToList().Contains("streetName"))
```
BenRamchandani, line 85: I think this is the same as
```suggestion
        if (results.Address.ContainsKey("streetName"))
```

BricksAndHearts.UnitTests/ServiceTests/Api/ApiServiceTestsBase.cs
```diff
@@ -0,0 +1,15 @@
+﻿using BricksAndHearts.Auth;
+using BricksAndHearts.Services;
+using FakeItEasy;
+using Microsoft.Extensions.Options;
+
+namespace BricksAndHearts.UnitTests.ServiceTests.Api;
+
+public class ApiServiceTestsBase
```
BenRamchandani, line 8: Don't feel like you need to split this into a separate file, fine to merge it all into ApiServiceTests

BricksAndHearts.UnitTests/ServiceTests/Api/ApiServiceTests.cs
```diff
@@ -0,0 +1,71 @@
...
+    {
+        // Arrange
+        string postalCode = "cb11dx";
+
+        // Act
+        // Uses the fake version of apiService because we don't want to make the API call
+        // Just want to check that the code works
+        string json = "some json string";
+        A.CallTo(() => _apiService.MakeApiRequestToAzureMaps(postalCode)).Returns(json);
+        string? responseBody = await _apiService.MakeApiRequestToAzureMaps(postalCode);
```
BenRamchandani, line 24: This isn't testing anything! You're setting up and then calling a fake.

We need to fake `HttpClient` in some manner. We'll need to inject it for that, so changes ApiService to accept HttpClient as a parameter, and add `builder.Services.AddHttpClient();` to `Program.cs`.

Unfortunately the `GetAsync` method on `HttpClient` isn't virtual, so we can't fake it directly. There's two approaches:

* Fake `HttpMessageHandler` and pass it to `HttpClient`
* Wrap `HttpClient` with our own wrapper with virtual methods and then fake that

I've done the first here:

```csharp
// Setup
var options = A.Fake<IOptions<AzureMapsOptions>>();
var logger = A.Fake<ILogger<ApiService>>();
var fakeMessageHandler = A.Fake<HttpMessageHandler>();
var fakeHttpClient = new HttpClient(fakeMessageHandler);
var underTest = new ApiService(logger, options, fakeHttpClient); // Note new argument

// Arrange
const string postalCode = "cb11dx";
var responseBody = await File.ReadAllTextAsync($"{AppDomain.CurrentDomain.BaseDirectory}/../../../ServiceTests/Api/AzureMapsApiResponse.json");
var response = new HttpResponseMessage
{
    StatusCode = HttpStatusCode.OK,
    Content = new StringContent(responseBody)
};
// Slightly icky because "SendAsync" is protected
A.CallTo(fakeMessageHandler).Where(c => c.Method.Name == "SendAsync").WithReturnType<Task<HttpResponseMessage>>().Returns(response);

// Act
var result = await underTest.MakeApiRequestToAzureMaps(postalCode);

// Assert
result.Should().NotBeNullOrEmpty("should not be null or empty");
```

web/Services/ApiService.cs
```diff
@@ -0,0 +1,100 @@
...
+    private readonly ILogger<ApiService> _logger;
+    private readonly IOptions<AzureMapsOptions> _options;
+
+    public ApiService(ILogger<ApiService> logger, IOptions<AzureMapsOptions> options)
+    {
+        _logger = logger;
+        _options = options;
+    }
+
+    public async Task<string> MakeApiRequestToAzureMaps(string postalCode)
```
BenRamchandani, line 27: This feels like it should be private, and is just public for testing. That isn't always wrong, but usually means there's a better way to organise things. For example we could have this function just deal with talking to Azure and have the autofilling logic somewhere else - either a new file or in PropertyService. (no need to change it, just for next time)

BenRamchandani (changes requested):
Well done working this all out!

I've left a few comments around cleaning the code up and the tests.

I've tried it out and found it works well, though there are a few things it would be nice to add at some point, but definitely separate tickets:

* Putting an invalid postcode in just throws an error
    * In the same vein we should handle the case where Azure maps errors and let them manually enter an address
* It would be nice if you could go back from the second step and change the postcode
* We should store the latitude and longitude of the postcode to make it easier to match tenants later

web/Services/ApiService.cs
```diff
@@ -0,0 +1,100 @@
+﻿using BricksAndHearts.Auth;
+using BricksAndHearts.ViewModels;
+using Microsoft.Extensions.Options;
+using Newtonsoft.Json;
+
+namespace BricksAndHearts.Services;
+
+public interface IApiService
```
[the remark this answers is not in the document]
Chestnut-lol: ok done

web/Services/ApiService.cs
Chestnut-lol: ok done! I've fixed up all the issues upon inspection.

web/Services/ApiService.cs
Chestnut-lol: ok I see :-) thank you!

web/ViewModels/PostcodeApiResponseViewModel.cs
Chestnut-lol: ok done

web/Services/ApiService.cs
Chestnut-lol: yep done

BricksAndHearts.UnitTests/ServiceTests/Api/ApiServiceTests.cs
Chestnut-lol: yeah sure - I have read and copied your code. Will definitely do something similar next time I come across stuff like this!

BricksAndHearts.UnitTests/ServiceTests/Api/ApiServiceTestsBase.cs
Chestnut-lol: Ok I've merged it all into one file.

web/Services/ApiService.cs
```diff
@@ -0,0 +1,100 @@
...
+    public ApiService(ILogger<ApiService> logger, IOptions<AzureMapsOptions> options)
+    {
+        _logger = logger;
+        _options = options;
+    }
+
+    public async Task<string> MakeApiRequestToAzureMaps(string postalCode)
+    {
+        string uri =
+            $"https://atlas.microsoft.com/search/address/structured/json?api-version=1.0&countryCode=GB&postalCode={postalCode}&subscription-key={_options.Value.SubscriptionKey}";
```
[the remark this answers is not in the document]
Chestnut-lol: Ok done.

Chestnut-lol:
I've changed the code and now if an invalid postcode is entered, it should just take us to the next page whereby the user can manually enter the addresses.

I will raise a ticket for "go back" options when filling in forms.

web/Auth/AzureMapsOptions.cs
```diff
@@ -0,0 +1,7 @@
+﻿namespace BricksAndHearts.Auth;
+
+public class AzureMapsOptions
+{
+    public const string AzureMaps = "AzureMaps";
+    public string SubscriptionKey { get; set; } = String.Empty;
```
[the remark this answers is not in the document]
BenRamchandani: Yep, don't put the actual key in, just put an empty string in appsettings.json, like for the Google ClientId and Secret.

It doesn't really matter, it's principal I guess that anyone should be able to figure out how to get it running with their own secrets without having access to our Keeper, and having an empty string in appsettings.json documents that the app needs a value there.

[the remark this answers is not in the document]
Chestnut-lol: ok done :-)

BenRamchandani (approved):
Looks good

BenRamchandani (approved):

PR #49, opened 2022-07-25
BNH-34 Add multiple pages to add property form

MaxTheComputerer:
The add new property form has been split over multiple pages. This has been designed so that it is very easy to add new pages/fields in the future. The design could also be (relatively) easily adapted for use elsewhere.

- When submitted, the first page of the form creates the record in the database, populated with the values from that page
- Each subsequent page will update that record with additional information, and redirect to the next page
- If the user navigates away from the form, they can resume adding their in-progress property by navigating back to the form, at any page
- Users can cancel an in-progress add, which will delete the partially-filled record from the database

To add new pages:
- Create a new partial view `_AddFormStep<num>.cshtml`, where `<num>` is the form page number
- This should contain all the HTML to go inside the form (i.e. all the `<input>`, `<label>`, etc. tags, but not `<form>` or the buttons)
- Modify `MaximumStep` in `AddNewPropertyViewModel` to be the new number of pages

image

image

image

image


web/Controllers/PropertyController.cs
```diff
@@ -0,0 +1,130 @@
...
+    }
+
+    [HttpPost("add/cancel")]
+    public ActionResult AddNewProperty_Cancel()
+    {
+        var landlordId = GetCurrentUser().LandlordId;
+        if (!landlordId.HasValue)
+        {
+            return StatusCode(403);
+        }
```
emilypatsko, line 114: Same

BricksAndHearts.UnitTests/ControllerTests/Landlord/LandlordControllerTests.cs
```diff
@@ -24,34 +21,8 @@ public void RegisterGet_CalledByUnregisteredUser_ReturnsRegisterViewWithEmail()
...
-        CreateUnregisteredUserInController(_underTest);
-        var formResultModel = A.Fake<PropertyViewModel>();
-
-        // Act
-        var result = _underTest.AddNewProperty(formResultModel) as StatusCodeResult;
-
-        // Assert
-        result!.StatusCode.Should().Be(403);
-    }
+    /*
```
emilypatsko, line 24: Can this commented out code be deleted (assuming it's just been moved elsewhere)?

BricksAndHearts.UnitTests/ControllerTests/Property/PropertyControllerTests.cs
```diff
@@ -0,0 +1,283 @@
...
+        // Act
+        var result = _underTest.AddNewProperty_Begin() as ViewResult;
+
+        // Assert
+        result!.ViewName.Should().Be("AddNewProperty");
+        result.Model.Should().BeOfType<AddNewPropertyViewModel>().Which.Step.Should().Be(1);
+    }
+
+    [Fact]
+    public void AddNewPropertyContinueGet_ReturnsViewAtStep()
```
emilypatsko, line 42: If there are multiple steps, you could turn this test into a `[Theory]` - this allows you to parameterise the test so you can ensure it returns the view at step 2/3/4 as appropriate

emilypatsko (changes requested):
Few small comments, but looking great so far

BricksAndHearts.UnitTests/ControllerTests/Landlord/LandlordControllerTests.cs
MaxTheComputerer: Yes! Thanks for spotting that 😅

BricksAndHearts.UnitTests/ControllerTests/Property/PropertyControllerTests.cs
MaxTheComputerer: That's a great idea, thank you!

web/Controllers/PropertyController.cs
```diff
@@ -0,0 +1,130 @@
...
+
+    [HttpGet("add/step/{step:int}")]
+    public ActionResult AddNewProperty_Continue([FromRoute] int step)
+    {
+        // Can replace with Role in future
+        var landlordId = GetCurrentUser().LandlordId;
+        if (!landlordId.HasValue)
+        {
+            return StatusCode(403);
+        }
```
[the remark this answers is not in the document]
MaxTheComputerer: Done! I've had to remove the unit test which checked for a 403 because this is now done by the `[Authorize(...)]` stuff, which can't be unit tested.

emilypatsko (approved):

emilypatsko (approved):

PR #51, opened 2022-07-25
BNH-54

SleepyUser:
Can now create landlords which are not assigned to a specific user


SleepyUser:
Forgot to do tests, will do them Tuesday (26th)

web/Controllers/LandlordController.cs
```diff
@@ -53,24 +59,50 @@ public async Task<ActionResult> RegisterPost([FromForm] LandlordProfileModel cre
...
                 _logger.LogWarning("User {UserId} already associated with landlord", user.Id);
                 TempData["FlashMessage"] = "Already registered!"; // This will be displayed on the Profile page
-                return Redirect(Url.Action("MyProfile")!);
+                if (!createModel.Unassigned)
+                {
+                    return Redirect(Url.Action("MyProfile")!);
+                }
+                else
+                {
+                    return RedirectToAction("LandlordList", "Admin");
```
BenRamchandani, line 104: This should never happen right? I wouldn't bother handling it, the original behaviour of redirecting to the profile with a flash message seems fine.

web/Migrations/BricksAndHeartsDbContextModelSnapshot.cs
```diff
@@ -65,7 +65,7 @@ protected override void BuildModel(ModelBuilder modelBuilder)
 
                     b.HasKey("Id");
 
-                    b.ToTable("Landlord");
+                    b.ToTable("Landlord", (string)null);
```
BenRamchandani, line 68: This shouldn't have changed without a migration, probably a merge issue

web/Services/LandlordService.cs
```diff
@@ -77,4 +78,31 @@ await using (var transaction = await _dbContext.Database.BeginTransactionAsync(I
...
+            LastName = createModel.LastName,
+            CompanyName = createModel.CompanyName,
+            Email = createModel.Email,
+            Phone = createModel.Phone,
+            LandlordStatus = createModel.LandlordStatus,
+            LandlordProvidedCharterStatus = createModel.LandlordProvidedCharterStatus
+        };
+
+        // Check there isn't already a Landlord with that email. Nothing depends on this currently, but it would probably mean the landlord is a duplicate
+        // This requires Serializable isolation, otherwise it will not lock any rows, and two racing registrations could create duplicate records
```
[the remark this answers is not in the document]
SleepyUser: I just found this while I was rebasing, so I'm halfway through this bit now :')

web/Migrations/BricksAndHeartsDbContextModelSnapshot.cs
SleepyUser: I think I changed this back during the rebase, it seemed weird there had been a change here

web/Controllers/LandlordController.cs
```diff
@@ -53,24 +59,50 @@ public async Task<ActionResult> RegisterPost([FromForm] LandlordProfileModel cre
...
-                _logger.LogInformation("Successfully created landlord for user {UserId}", user.Id);
-                return Redirect(Url.Action("MyProfile")!);
+                _logger.LogInformation("Successfully created landlord for user");
+                if (!createModel.Unassigned)
+                {
+                    return Redirect(Url.Action("MyProfile")!);
+                }
+                else
+                {
+                    return RedirectToAction("LandlordList", "Admin");
```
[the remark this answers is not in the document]
SleepyUser: Took me a little bit, but I managed to get this working

SleepyUser:
Tidied up commit history & rebased

web/Controllers/LandlordController.cs
```diff
@@ -44,25 +48,37 @@ public ActionResult RegisterGet()
...
 
         var user = GetCurrentUser();
+        ILandlordService.LandlordRegistrationResult result;
+        int? id = null;
+
+        if (createModel.Unassigned && user.IsAdmin)
+        {
+            var resultTuple = await _landlordService.RegisterLandlord(createModel);
+            result = resultTuple.result;
+            id = resultTuple.id;
```
BenRamchandani, line 61: Haven't tested it, but I think this works
```suggestion
            (result, id) = await _landlordService.RegisterLandlord(createModel);
```

web/Controllers/LandlordController.cs
```diff
@@ -44,25 +48,37 @@ public ActionResult RegisterGet()
...
 
         switch (result)
         {
             case ILandlordService.LandlordRegistrationResult.Success:
-                _logger.LogInformation("Successfully created landlord for user {UserId}", user.Id);
-                return Redirect(Url.Action("MyProfile")!);
+                _logger.LogInformation("Successfully created landlord for user");
+                if (!createModel.Unassigned)
+                    return Redirect(Url.Action("MyProfile")!);
+                //return RedirectToAction("LandlordList", "Admin");
```
BenRamchandani, line 75: Can delete this comment.

It's also easy to make mistakes doing `if` without braces. It's fine for things in one line like you've done below (e.g. `if (landlord == null) return StatusCode(404);`), but I'd use braces here.

BricksAndHearts.UnitTests/ControllerTests/Landlord/LandlordControllerTestsBase.cs
```diff
@@ -8,5 +8,6 @@ namespace BricksAndHearts.UnitTests.ControllerTests.Landlord;
 public class LandlordControllerTestsBase : ControllerTestsBase
 {
     public static readonly IPropertyService propertyService = A.Fake<IPropertyService>();
+    public static readonly ILandlordService landlordService = A.Fake<Services.LandlordService>();
     protected readonly LandlordController _underTest = new(null!, null!, new LandlordService(null!), propertyService);
```
BenRamchandani, line 12: We shouldn't be using static fakes, they should be fresh for each test. Can you add a constructor and set these up in there instead? Then you can use them in the test rather than creating them each time

BricksAndHearts.UnitTests/ControllerTests/Landlord/LandLordControllerUnassignedTests.cs
```diff
@@ -0,0 +1,45 @@
...
+using Xunit;
+
+namespace BricksAndHearts.UnitTests.ControllerTests.Landlord;
+
+public class LandLordControllerUnassignedTests : LandlordControllerTestsBase
+{
+    private static AdminController _adminController = A.Fake<AdminController>();
+
+    [Fact]
+    public async void AddNewUnassignedLandlord()
```
BenRamchandani, line 18: Would be good to have a test that non-admins cannot create unregistered landlords.

BenRamchandani (changes requested):

web/Controllers/LandlordController.cs
SleepyUser: This seems to work :)

BricksAndHearts.UnitTests/ControllerTests/Landlord/LandlordControllerTestsBase.cs
SleepyUser: I deleted these lines and all the tests still pass, , so it seems as though they're not required.
image

SleepyUser: All that was needed was to generate the fake in the UnderTest controller

SleepyUser: I've just seen there's more in the other files too :S

SleepyUser: I've rebased onto main because it already contained this fix

web/Controllers/LandlordController.cs
SleepyUser: Apologies, I've set up an automatic code format thing in rider and its a little overzealous. It has stopped all the "extra whitespace/newline" here comments though! ^^'

web/Controllers/LandlordController.cs
```diff
@@ -25,15 +25,21 @@ public LandlordController(ILogger<LandlordController> logger, ILandlordService l
...
         var currentUser = GetCurrentUser();
-        if (currentUser.LandlordId != null)
+        if (currentUser.LandlordId != null && !currentUser.IsAdmin)
         {
             _logger.LogWarning("User {UserId} is already registered, will redirect to profile", currentUser.Id);
             return Redirect(Url.Action("MyProfile")!);
         }
 
+        // If user is an admin, and there was a true createUnassigned param passed
+        if (currentUser.IsAdmin && createUnassigned)
```
BenRamchandani, line 38: Nit, this comment doesn't add anything

web/Controllers/LandlordController.cs
```diff
@@ -67,13 +75,15 @@ public async Task<ActionResult> RegisterPost([FromForm] LandlordProfileModel cre
                                  + $"Email: {createModel.Email}" + "\n"
                                  + $"Phone: {createModel.Phone}" + "\n";
                 _mailService.SendMsg(msgBody);
-
-                return Redirect(Url.Action("MyProfile")!);
+                if (!createModel.Unassigned)
+                    return Redirect(Url.Action("MyProfile")!);
+                //return RedirectToAction("LandlordList", "Admin");
```
BenRamchandani, line 80: Delete commented out code

web/Services/LandlordService.cs
```diff
@@ -79,25 +81,66 @@ await using (var transaction = await _dbContext.Database.BeginTransactionAsync(I
...
+            LandlordStatus = createModel.LandlordStatus,
+            LandlordProvidedCharterStatus = createModel.LandlordProvidedCharterStatus
+        };
+        await using (var transaction = await _dbContext.Database.BeginTransactionAsync(IsolationLevel.Serializable))
+        {
+            // Check there isn't already a Landlord with that email. Nothing depends on this currently, but it would probably mean the landlord is a duplicate
+            // This requires Serializable isolation, otherwise it will not lock any rows, and two racing registrations could create duplicate records
+            if (await _dbContext.Landlords.AnyAsync(l => l.Email == createModel.Email))
+                return (ILandlordService.LandlordRegistrationResult.ErrorLandlordEmailAlreadyRegistered,
+                    await GetLandlordIdIfExistsFromModel(dbModel) ?? 0);
```
BenRamchandani, line 104: So this returns the id of the landlord with the same email on an email error? That seems odd and I don't think it's necessary.

I'd return null for the id here (need to change the return type to `int? id`), it's a bit clearer than returning 0.

web/Services/LandlordService.cs
```diff
@@ -79,25 +81,66 @@ await using (var transaction = await _dbContext.Database.BeginTransactionAsync(I
...
+
+            // Insert the landlord and call SaveChanges
+            // Entity Framework will insert the record and populate dbModel.Id with the new record's id
+            _dbContext.Landlords.Add(dbModel);
+            await _dbContext.SaveChangesAsync();
+            await transaction.CommitAsync();
+        }
+
+        var id = await GetLandlordIdIfExistsFromModel(dbModel) ?? 0;
+        return (ILandlordService.LandlordRegistrationResult.Success, id);
```
BenRamchandani, line 114: Can just do `dbModel.Id`

web/Services/LandlordService.cs
```diff
@@ -79,25 +81,66 @@ await using (var transaction = await _dbContext.Database.BeginTransactionAsync(I
         return ILandlordService.LandlordRegistrationResult.Success;
     }
 
+    public async Task<(ILandlordService.LandlordRegistrationResult result, int id)> RegisterLandlord(
```
BenRamchandani, line 84: I'd be tempted to return the whole `LandlordDbModel` on success, rather than just the id. It's slightly clearer what's going on that way.

I think also change the method above to have the same return type, for consistencies sake.

BenRamchandani (changes requested):

web/Controllers/LandlordController.cs
SleepyUser: Fixed it :) told it to keep embedded blocks as they are in the code

web/Controllers/LandlordController.cs
SleepyUser: Apologies, I pushed to save my work midway through fixing the ones above ^^' I didn't realise it would be re-reviewed without pressing the button.

SleepyUser: If I turn it into a draft in future, will it retain your comments?

BenRamchandani: Yeah I believe so

BricksAndHearts.UnitTests/ControllerTests/Landlord/LandLordControllerUnassignedTests.cs
```diff
@@ -0,0 +1,70 @@
...
+        result.Should().BeOfType<RedirectToActionResult>();
+        result.Should().NotBeNull();
+        if (result != null)
+        {
+            result.ActionName.Should().BeEquivalentTo("Profile");
+            result.ControllerName.Should().BeEquivalentTo("Landlord");
+            Assert.False(result.RouteValues.IsNullOrEmpty());
+            result.RouteValues?["id"].Should().Be(returnedLandlord.Id);
+        }
+    }
```
BenRamchandani, line 43: Can tidy this up a bit:

* No need to use fakes for simple objects if you don't need to mock anything
* `.DoesNothing();` is the default behaviour of FakeItEasy it seems, so we can omit that setup
* Setting up mocks (`A.CallTo`) normally sits in the Arrange section
* Can be a bit more liberal with the assume not null `!` operator in tests, no need to check separately

```suggestion
    [Fact]
    public async void
        RegisterLandlordWithNoAssignedUser_WhenAttemptedByAdmin_CreatesNewUnassignedLandlord_AndRedirectsToTheirProfile()
    {
        // Arrange
        var adminUser = CreateAdminUser();
        MakeUserPrincipalInController(adminUser, UnderTest);

        var returnedLandlord = new LandlordDbModel
        {
            Id = 5
        };
        var formResultModel = new LandlordProfileModel
        {
            Unassigned = true
        };
        A.CallTo(() => LandlordService.RegisterLandlord(formResultModel))
            .Returns((ILandlordService.LandlordRegistrationResult.Success, returnedLandlord));

        // Act
        var result = await UnderTest.RegisterPost(formResultModel) as RedirectToActionResult;

        // Assert
        result.Should().BeOfType<RedirectToActionResult>();
        result.Should().NotBeNull();
        result!.ActionName.Should().BeEquivalentTo("Profile");
        result.ControllerName.Should().BeEquivalentTo("Landlord");
        result.RouteValues!["id"].Should().Be(returnedLandlord.Id);
    }
```

web/Controllers/LandlordController.cs
```diff
@@ -51,29 +57,47 @@ public async Task<ActionResult> RegisterPost([FromForm] LandlordProfileModel cre
...
 
-        var result = await _landlordService.RegisterLandlordWithUser(createModel, user);
+        if (createModel.Unassigned && user.IsAdmin)
+        {
+            (result, landlord) = await _landlordService.RegisterLandlord(createModel);
+        }
+        else
+        {
+            return StatusCode(403);
+        }
```
BenRamchandani, line 70: Still need to handle the regular registration case!

web/Controllers/LandlordController.cs
```diff
@@ -51,29 +57,47 @@ public async Task<ActionResult> RegisterPost([FromForm] LandlordProfileModel cre
...
+                if (!createModel.Unassigned)
+                {
+                    return Redirect(Url.Action("MyProfile")!);
+                }
 
-                return Redirect(Url.Action("MyProfile")!);
+                if (landlord == null)
+                {
+                    return StatusCode(500);
+                }
```
BenRamchandani, line 93: I'd rather just let it throw an exception on the next line, that's easier to debug than a naked 500 response

web/Services/LandlordService.cs
```diff
@@ -75,30 +83,71 @@ await using (var transaction = await _dbContext.Database.BeginTransactionAsync(I
...
+            LandlordStatus = createModel.LandlordStatus,
+            LandlordProvidedCharterStatus = createModel.LandlordProvidedCharterStatus
+        };
+        await using (var transaction = await _dbContext.Database.BeginTransactionAsync(IsolationLevel.Serializable))
+        {
+            // Check there isn't already a Landlord with that email. Nothing depends on this currently, but it would probably mean the landlord is a duplicate
+            // This requires Serializable isolation, otherwise it will not lock any rows, and two racing registrations could create duplicate records
+            if (await _dbContext.Landlords.AnyAsync(l => l.Email == createModel.Email))
+            {
+                return (ILandlordService.LandlordRegistrationResult.ErrorLandlordEmailAlreadyRegistered, dbModel);
```
BenRamchandani, line 133: nitpick, but don't need to return the result

```suggestion
                return (ILandlordService.LandlordRegistrationResult.ErrorLandlordEmailAlreadyRegistered, null);
```

BenRamchandani (changes requested):
Few things to update.

Don't worry now, but next time if you're going to make unrelated formatting changes raise a separate PR - it's then easy to review that and get it merged quickly before it goes out of date, and means there's less to review in the PR with changes in.

web/Controllers/LandlordController.cs
SleepyUser: Oh dear, there isn't a test for it is there? That's why they all still passed

BricksAndHearts.UnitTests/ControllerTests/Landlord/LandLordControllerUnassignedTests.cs
SleepyUser: so it's okay to leave the null checks out as they'd cause exceptions only during testing anyway?

SleepyUser:
I've added a test for signing up as a non-landlord, so removing that case doesn't happen again 🤦‍♂️

BricksAndHearts.UnitTests/ControllerTests/Landlord/LandLordControllerUnassignedTests.cs
BenRamchandani: Yeah. Having `result.Should().NotBeNull();` will give you a slightly nicer error message than a null pointer exception on the next line, but I don't think it's worth the extra code, being able to quickly scan a test and work out what it's doing is helpful.

In non-test code it's good to be explicit about things like that, but it doesn't really matter in tests.

BricksAndHearts.UnitTests/ControllerTests/Landlord/LandlordControllerTests.cs
```diff
@@ -25,6 +25,27 @@ public void RegisterGet_CalledByUnregisteredUser_ReturnsRegisterViewWithEmail()
...
+        A.CallTo(() => LandlordService.RegisterLandlord(formResultModel, unregisteredUser))
+            .Returns((ILandlordService.LandlordRegistrationResult.Success, returnedLandlord));
+
+        // Act
+        var result = await UnderTest.RegisterPost(formResultModel) as RedirectToActionResult;
+
+        // Assert
+        result.Should().BeOfType<RedirectToActionResult>();
+        result.Should().NotBeNull();
+        result!.ActionName.Should().BeEquivalentTo("MyProfile");
```
BenRamchandani, line 46: Could also assert that the call to the service has happened

BenRamchandani (dismissed):

BricksAndHearts.UnitTests/ControllerTests/Landlord/LandlordControllerTests.cs
SleepyUser: I was just thinking that as I reviewed Claire's code and saw she had 😅

SleepyUser: Added the lines in to check there was or wasn't calls to specific services :)

BenRamchandani (dismissed):
Looks good, and thanks for adding that test

SleepyUser:
Had to rebase it, it said there were conflicts again 😱

BenRamchandani (approved):

PR #52, opened 2022-07-26
Bnh 44

cec200wire:
LandlordList page defaults to displaying a list of all landlords. There is then an option to filter the list to display 'Unapproved', 'Approved', or 'All'.
This is written as a Get Request, because using a Post Request caused the 'confirm form resubmission page' to inconveniently appear if, after clicking on a landlord to see their profile, the user uses the browser back button to return to the Landlord List.

### DIFF
--- a/BricksAndHearts.UnitTests/ControllerTests/Admin/AdminControllerTests.cs
+++ b/BricksAndHearts.UnitTests/ControllerTests/Admin/AdminControllerTests.cs
@@ -37,7 +37,7 @@ public class AdminControllerTests : AdminControllerTestsBase
         var result = await UnderTest.LandlordList() as ViewResult;
 
         // Assert
-        A.CallTo(() => AdminService.GetLandlordDisplayList("All")).MustHaveHappened();
+        A.CallTo(() => AdminService.GetLandlordDisplayList("")).MustHaveHappened();
         result!.ViewData.Model.Should().BeOfType<LandlordListModel?>();
     }
 }
--- a/BricksAndHearts.UnitTests/ControllerTests/Admin/AdminControllerTests.cs
+++ b/BricksAndHearts.UnitTests/ControllerTests/Admin/AdminControllerTests.cs
@@ -27,7 +27,7 @@ public class AdminControllerTests : AdminControllerTestsBase
     }
 
     [Fact]
-    public async void LandlordList_WhenCalled_CallsGetUnapprovedLandlordsAndReturnsLandlordListView()
+    public async void LandlordList_WhenCalled_CallsGetLandlordListAndReturnsLandlordListView()
     {
         // Arrange
         var adminUser = CreateAdminUser();
@@ -37,7 +37,7 @@ public class AdminControllerTests : AdminControllerTestsBase
         var result = await UnderTest.LandlordList() as ViewResult;
 
         // Assert
-        A.CallTo(() => AdminService.GetUnapprovedLandlords()).MustHaveHappened();
+        A.CallTo(() => AdminService.GetLandlordDisplayList("All")).MustHaveHappened();
         result!.ViewData.Model.Should().BeOfType<LandlordListModel?>();
     }
 }
--- a/web/Controllers/AdminController.cs
+++ b/web/Controllers/AdminController.cs
@@ -1,7 +1,6 @@
 #region
 
 using BricksAndHearts.Auth;
-using BricksAndHearts.Database;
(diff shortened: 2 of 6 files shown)


BricksAndHearts.UnitTests/ControllerTests/Admin/AdminControllerTests.cs
```diff
@@ -37,7 +37,7 @@ public async void LandlordList_WhenCalled_CallsGetUnapprovedLandlordsAndReturnsL
         var result = await _underTest.LandlordList() as ViewResult;
 
         // Assert
-        A.CallTo(() => adminService.GetUnapprovedLandlords()).MustHaveHappened();
+        A.CallTo(() => adminService.GetLandlordDisplayList("")).MustHaveHappened();
```
emilypatsko, line 40: You could mock this out above in the arrange section and make it return a specific list, then assert that the view model contains the correct data

web/Controllers/AdminController.cs
```diff
@@ -80,13 +80,13 @@ public async Task<IActionResult> AdminList()
         AdminListModel adminListModel = new AdminListModel(adminLists.CurrentAdmins, adminLists.PendingAdmins);
         return View(adminListModel);
     }
-    
+
     [Authorize(Roles="Admin")]
     [HttpGet]
-    public async Task<IActionResult> LandlordList()
+    public async Task<IActionResult> LandlordList(string? approvalStatus = "")
```
emilypatsko, line 86: For consistency I'd call this method something like `GetLandlordList`

web/Services/AdminService.cs
```diff
@@ -44,18 +44,23 @@ private async Task<List<UserDbModel>> GetCurrentAdmins()
 
     private async Task<List<UserDbModel>> GetPendingAdmins()
     {
-        List<UserDbModel> PendingAdmins = await _dbContext.Users.Where(u => u.IsAdmin == false && u.HasRequestedAdmin).ToListAsync();
+        List<UserDbModel> PendingAdmins =
+            await _dbContext.Users.Where(u => u.IsAdmin == false && u.HasRequestedAdmin).ToListAsync();
```
emilypatsko, line 48: I'd indent this differently
```suggestion
        List<UserDbModel> PendingAdmins = await _dbContext.Users.Where(u => 
            u.IsAdmin == false && 
            u.HasRequestedAdmin)
            .ToListAsync();
```

emilypatsko (changes requested):

BricksAndHearts.UnitTests/ControllerTests/Admin/AdminControllerTests.cs
cec200wire: Is there an advantage to doing it that way? I might be misunderstanding, but it seems as though that would only confirm that the fake service is called and that the return from that service is registered by the controller - both parts of which are checked by the current test  - but would require a lot more set-up.

emilypatsko: It can be useful to check that the data returned from the service is correctly mapped to the view model.

web/Services/AdminService.cs
```diff
@@ -48,21 +48,23 @@ private async Task<List<UserDbModel>> GetCurrentAdmins()
 
     private async Task<List<UserDbModel>> GetPendingAdmins()
     {
-        var pendingAdmins =
-            await _dbContext.Users.Where(u => u.IsAdmin == false && u.HasRequestedAdmin).ToListAsync();
-        return pendingAdmins;
+        List<UserDbModel> PendingAdmins = await _dbContext.Users.Where(u => u.IsAdmin == false && u.HasRequestedAdmin).ToListAsync();
```
emilypatsko, line 51: Do you not get a warning on this line about `PendingAdmins` not meeting the naming convention?

emilypatsko (changes requested):
Small change. Could you also add a screenshot or similar to this PR to show what this functionality looks like?

web/Services/AdminService.cs
cec200wire: Yes, I'm not quite sure how that slipped through before.

cec200wire:
image
image

emilypatsko (commented):
You have a typo in your last commit 😛 quick rebase to fix it. But then this is good to go

emilypatsko (approved):

PR #55, opened 2022-07-26
BNH-61 Unit tests for database methods

MaxTheComputerer:
## Summary of changes

- Created a framework for unit testing our database operations found in the Service classes, following [this guide](https://docs.microsoft.com/en-us/ef/core/testing/testing-with-the-database) from the Entity Framework docs
- This uses a new `BricksAndHeartsTest` database
- This can test read and write operations, and uses a non-committing transaction to maintain isolation between tests
- The pull request GitHub Action has been updated to now run an instance of SQL Server, which is used when these tests are run for pull requests
- Added a small number of example tests for `AdminService` to demonstrate how this should be used

### DIFF
--- a/.github/workflows/test_on_push_bricks-and-hearts.yml
+++ b/.github/workflows/test_on_push_bricks-and-hearts.yml
@@ -12,6 +12,21 @@ jobs:
   build:
     name: Build and run tests
     runs-on: ubuntu-latest
+    
+    services:
+      sqlserver:
+        image: mcr.microsoft.com/mssql/server:2019-latest
+        env:
+          ACCEPT_EULA: Y
+          SA_PASSWORD: JdMsKZPBBA8kVFXVrj8d
+        ports:
+          - 1433:1433
+        options: >-
+          --health-cmd "/opt/mssql-tools/bin/sqlcmd -S localhost -U sa -P 'JdMsKZPBBA8kVFXVrj8d' -Q 'SELECT 1' || exit 1"
+          --health-interval 10s
+          --health-timeout 3s
+          --health-retries 10
+          --health-start-period 10s
 
     steps:
       - uses: actions/checkout@v2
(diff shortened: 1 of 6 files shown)


MaxTheComputerer:
@BenRamchandani 
> My only comment is we should make sure we separate out the tests in some way, whether that's in a folder, or with xUnit's Trait annotation, or a separate project.
Do you mean separating these database tests from the normal unit tests?

> I thought SQL Server supported nested transactions, not sure why it wouldn't work
Yeah the guide I was following had a complicated special case for transactional tests, but looking closer it seems that's because EF can't guarantee all databases support nested transactions. If SQL Server does support them then it should work fine 🙂 

I'll have a look at the CI stuff, might use this draft PR to test it out. Thanks!

MaxTheComputerer:
Entity Framework doesn't seem to support nested transactions, giving the exception: `System.InvalidOperationException: The connection is already in a transaction and cannot participate in another transaction`. Discussion with Ben concluded that we should not worry about transactional tests for now.

.github/workflows/test_on_push_bricks-and-hearts.yml
```diff
@@ -12,6 +12,21 @@ jobs:
...
+        image: mcr.microsoft.com/mssql/server:2019-latest
+        env:
+          ACCEPT_EULA: Y
+          SA_PASSWORD: JdMsKZPBBA8kVFXVrj8d
+        ports:
+          - 1433:1433
+        options: >-
+          --health-cmd "/opt/mssql-tools/bin/sqlcmd -S localhost -U sa -P 'JdMsKZPBBA8kVFXVrj8d' -Q 'SELECT 1' || exit 1"
+          --health-interval 10s
+          --health-timeout 3s
```
BenRamchandani, line 27: I'd probably lower the interval a bit, and up the retries, no need to wait 10s

BenRamchandani (dismissed):
Looks great :)

.github/workflows/test_on_push_bricks-and-hearts.yml
```diff
@@ -12,6 +12,21 @@ jobs:
   build:
     name: Build and run tests
     runs-on: ubuntu-latest
+    
+    services:
+      sqlserver:
+        image: mcr.microsoft.com/mssql/server:2019-latest
```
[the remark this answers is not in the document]
MaxTheComputerer: Yeah I came across it when looking at how other people have used SQL Server from GitHub actions. Seems like a nice way of avoiding executing `docker` commands yourself!

.github/workflows/test_on_push_bricks-and-hearts.yml
MaxTheComputerer: I've adjusted the parameters, hopefully this is more reasonable

BenRamchandani (approved):

PR #56, opened 2022-07-26
Bnh 46

blancamc00:
image
Admins can see a list of all properties, each with a button that takes to their landlord's profile. 
Database was updated to include in the properties the id of the user who rented it. If it is null, the property is available. The table can be sorted by availability, which puts available properties first, or by rent in ascending or descending order.
The table is paginated with 10 properties in each tab and previous and next buttons.


web/Migrations/20220726082722_PropertyContainsUserWhoRented.cs
```diff
@@ -0,0 +1,188 @@
...
+    {
+        protected override void Up(MigrationBuilder migrationBuilder)
+        {
+            migrationBuilder.AlterColumn<string>(
+                name: "TownOrCity",
+                table: "Property",
+                type: "nvarchar(max)",
+                nullable: true,
+                oldClrType: typeof(string),
+                oldType: "nvarchar(max)");
```
BenRamchandani, line 18: I don't think all of these should be here, maybe need to regenerate this after rebasing + applying migrations?

web/Controllers/AdminController.cs
```diff
@@ -95,4 +95,36 @@ private void LoggerAlreadyAdminWarning(ILogger logger, BricksAndHeartsUser user)
...
+    {
+        var landlord = await _adminService.GetLandlordDbModelFromId(landlordId);
+        if (landlord == null)
+        {
+            return StatusCode(404);
+        }
+
+        var LandlordViewModel = LandlordProfileModel.FromDbModel(landlord);
+        return View("Profile", LandlordViewModel);
+    }
```
BenRamchandani, line 116: Can we link to the regular landlord profile on the LandlordController? It would be nice if we can avoid duplicating this and Profile.cshtml

web/Views/Admin/AllProperties.cshtml
```diff
@@ -0,0 +1,102 @@
...
+<h2>All properties</h2>
+
+<div>
+    <div>
+        <div asp-validation-summary="All" class="text-danger"></div>
+            @{
+                var totalRow = 0;
+                var propPerTab = 10;
+            }
+            @for (int tab = 0; tab <= Model.Properties.Count / propPerTab; tab++)
```
BenRamchandani, line 16: This is paginating it client side, which is to say all the properties are sent to the browser, and then we're using javascript to only show some of them at once.

Server side pagination would be to only send the current page to the browser, and then make another request when we want a different page.

Let me know if that doesn't make sense ^

Client side pagination works well for up to a few hundred records since switching tabs is instant, but past that we should use server side. If we're really going to have thousands of properties like ChangeAhead say, then we'll want this to be paginated server side at some point. You can do that now if you'd like, but otherwise lets merge this and raise another ticket to come back to it.

BenRamchandani (changes requested):

blancamc00:
Server side pagination

web/BricksAndHearts.csproj
```diff
@@ -9,8 +9,9 @@
 
     <ItemGroup>
         <PackageReference Include="Azure.Storage.Blobs" Version="12.13.0" />
-        <PackageReference Include="Microsoft.ApplicationInsights.AspNetCore" Version="2.21.0" />
         <PackageReference Include="MailKit" Version="3.3.0" />
+        <PackageReference Include="Microsoft.ApplicationInsights.AspNetCore" Version="2.21.0" />
+        <PackageReference Include="jQuery" Version="3.6.0" />
```
BenRamchandani, line 14: Are we using this? I don't think we need jQuery as a C# package, we have a copy in `wwwroot/js/lib`

BenRamchandani (changes requested):
This looks good, I've left some comments on things we can simplify/deduplicate

BricksAndHearts.UnitTests/BricksAndHearts.UnitTests.csproj
```diff
@@ -10,6 +10,8 @@
     <ItemGroup>
         <PackageReference Include="FakeItEasy" Version="7.3.1" />
         <PackageReference Include="FluentAssertions" Version="6.7.0" />
+        <PackageReference Include="jQuery" Version="3.6.0" />
```
BenRamchandani, line 13: Can remove from here too

BenRamchandani (approved):

BenRamchandani (approved):

PR #57, opened 2022-07-26
Bnh 55

Chestnut-lol:
Added a button at the bottom of landlord profile page to get the invite link for this landlord:
image
No invite link is generated if the landlord is already linked to an existing user in the database:
image
If landlord already has an invite link, will display the link with the option to re-new
image
Click on it will generate a new invite link for the landlord:
image


Chestnut-lol:
I will need to add unit tests and will open this again

Chestnut-lol:
Most of the code is front-end and database interactions so I didn't add any tests to it. I manually test it on the front-end and it all seems to work fine.

web/Controllers/AdminController.cs
```diff
@@ -89,4 +89,58 @@ public async Task<IActionResult> LandlordList()
         landlordListModel.UnapprovedLandlords = await _adminService.GetUnapprovedLandlords();
         return View(landlordListModel);
     }
+
+    [Authorize(Roles = "Admin")]
+    [HttpPost]
+    public Task<ActionResult> GetInviteLink(int landlordId)
```
BenRamchandani, line 95: Methods that return `Task` should almost always be async. There's a pretty good summary of dos and don'ts here: https://github.com/davidfowl/AspNetCoreDiagnosticScenarios/blob/master/AsyncGuidance.md and I'm happy to chat through async at some point if it doesn't make sense.

web/Migrations/20220726103731_AddInviteLinkToLandlord.cs
```diff
@@ -0,0 +1,188 @@
...
+    {
+        protected override void Up(MigrationBuilder migrationBuilder)
+        {
+            migrationBuilder.AlterColumn<string>(
+                name: "TownOrCity",
+                table: "Property",
+                type: "nvarchar(max)",
+                nullable: true,
+                oldClrType: typeof(string),
+                oldType: "nvarchar(max)");
```
BenRamchandani, line 18: These extra alters shouldn't be here. Possibly there was a missing migration on main at some point? Will hopefully be fixed by rebasing and recreating this migration

web/Services/AdminService.cs
```diff
@@ -44,18 +48,68 @@ private async Task<List<UserDbModel>> GetCurrentAdmins()
...
+            await _dbContext.Landlords.Where(u => u.CharterApproved == false).ToListAsync();
         return UnapprovedLandlords;
     }
+
+    public UserDbModel? FindUserByLandlordId(int landlordId)
+    {
+        return _dbContext.Users.SingleOrDefault(u => u.LandlordId == landlordId);
+    }
+    
+    public string FindExistingInviteLink(int landlordId)
```
BenRamchandani, line 73: It might be slightly more intuitive to return `null` in the case where there's no link, then the return type here would be `string?`

web/Views/Landlord/Profile.cshtml
```diff
@@ -15,6 +15,25 @@
             </div>
         }
     }
+    @{
+        if (TempData["ToastMessage"] != null)
```
BenRamchandani, line 19: Might be worth being a bit more explicit rather than "ToastMessage", could call it e.g. "InviteLinkMessage"

web/Views/Landlord/Profile.cshtml
```diff
@@ -72,4 +91,8 @@
         <p>Charter status: Pending</p>
     }
     
+    @using (Html.BeginForm("GetInviteLink", "Admin", new { landlordId = Model.LandlordId }, FormMethod.Post))
```
BenRamchandani, line 94: Can you make sure this is only visible to admins?

BenRamchandani (changes requested):
A few small comments, but this looks good 🙂

web/Controllers/AdminController.cs
Chestnut-lol: yeah sure I'll look over the document
Rider doesn't seem to be very happy when I change it to async as there is no await calls inside the function
Should I change some of the functions in the service to async too and await for them in the controller?

Chestnut-lol: Rider doesn't seem to be very happy when I just add async, since there is no await inside the call.

Instead I changed it from Task<ActionResult> to just ActionResult and it seems to be happy.

Is it ok to just do that?

web/Services/AdminService.cs
Chestnut-lol: yeah sure done!

web/Views/Landlord/Profile.cshtml
Chestnut-lol: ok done

web/Views/Landlord/Profile.cshtml
Chestnut-lol: Yeah ok done.

web/Migrations/20220726103731_AddInviteLinkToLandlord.cs
Chestnut-lol: yes I've rebased onto main and it is now all clean.

web/Controllers/AdminController.cs
BenRamchandani: Yep, just having "normal" non-async code is fine, we haven't really standardised on one or the other.

BenRamchandani (approved):
Looks good 🙂

emilypatsko (approved):

PR #59, opened 2022-07-27
BNH-64 property table styling

TomHanmer29:
Some minimal styling, added postcode to properties and removed extra commas between whitespace depending on if a line 2 and 3 are present in the address


emilypatsko:
Could you add a screenshot? 😊

emilypatsko (approved):

TomHanmer29:
image
Here you go!

PR #63, opened 2022-07-28
BNH-71 bug: Fix up bugs for postcode autocomplete.

Chestnut-lol:


MaxTheComputerer (approved):
Looks good to me!

Chestnut-lol:
I fixed up all bugs raised during testing the previous ticket (sending emails). All things should (hopefully) be working fine now.

One problem is that the API request (i.e. autofilling the postcode) seems to be taking 5 to 6 seconds. It was faster before I did the bugfix, but I didn't change anything with the api request so I don't know if it's just a temporary internet connection problem, or if I accidentally changed something to make it slower...

web/Controllers/PropertyController.cs
```diff
@@ -57,36 +58,46 @@ public async Task<IActionResult> AddNewProperty_Continue([FromRoute] int step, [
...
+        {
+            if (newPropertyModel.Address.AddressLine1 == null || newPropertyModel.Address.Postcode == null)
+            {
+                // Address line 1 and postcode are the minimum information we need to create a new record
+                return View("AddNewProperty",
+                    new AddNewPropertyViewModel { Step = step, Property = newPropertyModel });
+            }
+
+            PropertyViewModel resultModel = await _azureMapsApiService.AutofillAddress(newPropertyModel);
+            newPropertyModel = resultModel;
```
BenRamchandani, line 78: The method mutates it anyway, so just this should work

```suggestion
            await _azureMapsApiService.AutofillAddress(newPropertyModel);
```

Might be worth changing `AutofillAddress` to just return a `Task` to make that clear

BenRamchandani (commented):
One tiny thing, but this looks good

BenRamchandani (approved):

PR #65, opened 2022-07-28
Bnh 56 get invite link working

Chestnut-lol:
Now if enter an invite link (/invite/link=...) will take to a page which tells you either:
1. The link does not work "Ooops the link does not work"
2. The link does work and the user has the option to accept the invite

If the user clicks "Accept invite", one of the following will happen:
1. If the user is already linked to a landlord, nothing will happen on the backend and the user will be taken to their _own_ landlord profile
2. If the user is not linked to a landlord, they will be linked to this landlord (with the invite link), and the invite link will no longer work
3. If the invite link is somehow invalid (e.g. if the user waits too long to click accept invite, during the time someone else accepted the invite elsewhere), then the user will be taken again to the page telling them "Ooops the link does not work"


web/Views/Landlord/Invite.cshtml
```diff
@@ -0,0 +1,19 @@
+@{
+    ViewData["Title"] = "Accept Invite";
+}
+
+@if (TempData["InviteLinkToAccept"] != null)
```
BenRamchandani, line 5: Can you make a view model for this rather than using TempData?

TempData is great for stuff like flash messages that you want to display on another page, but they're not typed so it's better to use a view model for stuff that you'll always display.

web/Controllers/LandlordController.cs
```diff
@@ -141,4 +140,44 @@ public IActionResult ViewProperties()
         var listOfProperties = databaseResult.Select(PropertyViewModel.FromDbModel).ToList();
         return View("Properties", new PropertiesDashboardViewModel(listOfProperties));
     }
+
+    [HttpGet]
+    [Route("/invite/link={inviteLink}")]
+    public ActionResult Invite(string inviteLink)
```
BenRamchandani, line 146: This will work, but it's an unusual format for a URL. I think do one of

Have it in a query string, `/invite?link=123abc`

```cs
    [Route("/invite")]
    public ActionResult Invite([FromQuery] string inviteLink)
```

or have it as part of the path with no `=`, so `/invite/123abc`

```cs
    [Route("/invite/{inviteLink}")]
    public ActionResult Invite([FromRoute] string inviteLink)
```

web/Services/LandlordService.cs
```diff
@@ -100,4 +113,49 @@ public async Task ApproveLandlord(int landlordId,  BricksAndHeartsUser user)
...
+        return _dbContext.Landlords.SingleOrDefault(l => l.InviteLink == inviteLink);
+    }
+    
+    // Link existing landlord with user
+    public async Task<ILandlordService.LinkUserWithLandlordResult> LinkExistingLandlordWithUser(
+        string inviteLink,
+        BricksAndHeartsUser user)
+    {
+        // We want to atomically update multiple records (insert a landlord, then set the user's landlord id), so first start a transaction
+        var transaction = await _dbContext.Database.BeginTransactionAsync();
```
BenRamchandani, line 128:
```suggestion
        var transaction = await _dbContext.Database.BeginTransactionAsync(IsolationLevel.RepeatableRead);
```

web/Services/LandlordService.cs
```diff
@@ -100,4 +113,49 @@ public async Task ApproveLandlord(int landlordId,  BricksAndHeartsUser user)
...
+            await _dbContext.SaveChangesAsync();
+
+            // Now we can update the user record too
+            userRecord.LandlordId = landlord.Id;
+            userRecord.Landlord = landlord;
+            user.LandlordId = landlord.Id;
+            await _dbContext.SaveChangesAsync();
+
+            // and finally commit
+            await transaction.CommitAsync();
```
BenRamchandani, line 157: This is mostly my fault for misleading comments/code elsewhere. We don't need two calls to SaveChanges here. We do still need the transaction to make sure someone else isn't using the invite link at the same time as us - and we need `REPEATABLE READ` for that, since we need to lock the landlord record.

In the database the only way to link landlords to users is to set user.LandlordId. Entity Framework should translate all of these into the same thing, so we only need one

* `landlord.User = userRecord;`
* `userRecord.LandlordId = landlord.Id;`
* `user.LandlordId = landlord.Id;`

```suggestion
            landlord.InviteLink = null;
            userRecord.LandlordId = landlord.Id;
            await _dbContext.SaveChangesAsync();
            await transaction.CommitAsync();

            user.LandlordId = landlord.Id; // Update the in memory user object that `GetCurrentUser()` returns
```

BricksAndHearts.UnitTests/ControllerTests/Landlord/LandlordControllerTests.cs
```diff
@@ -37,4 +42,47 @@ public async void ApproveCharter_CallsApproveLandlord()
...
-}
+
+    [Fact]
+    public async void TieUserWithLandlord_WithNonExistentLink_ReturnDirectToInvite()
+    {
+        // Arrange 
+        var landlordUser = CreateLandlordUser();
+        string inviteLink = "11111";
+        var landlordService = A.Fake<ILandlordService>();
+        var logger = A.Fake<ILogger<LandlordController>>();
```
BenRamchandani, line 53: This setup stuff should be done in the class now, might need to rebase

BenRamchandani (changes requested):

web/Views/Landlord/Invite.cshtml
Chestnut-lol: yes sure ok

web/Controllers/LandlordController.cs
Chestnut-lol: yeah ok I did the latter one.

web/Services/LandlordService.cs
Chestnut-lol: ok :-)

web/Services/LandlordService.cs
Chestnut-lol: ok :-)

BricksAndHearts.UnitTests/ControllerTests/Landlord/LandlordControllerTests.cs
Chestnut-lol: yep

web/Services/LandlordService.cs
Chestnut-lol: Hey Ben, just a quick question - I'm trying to write database tests, but so far it doesn't quite work if the method involves a transaction. My code for this method uses a transaction because I reckon that the method resembles RegisterLandlordWithUser much, and so I wrote my code in a similar way to RegisterLandlordWithUser. Do you think that I should modify it in a way so that it doesn't use transactions, so that all of it can actually be tested?

BenRamchandani: Yeah tests with transactions don't work at the moment because of how it's set up 🙁 

I think just don't worry about testing this, you're absolutely right to use a transaction, but it's not worth the effort of setting up testing for it - the tests use transactions to keep each test isolated from all the others, but we can't nest them.

BricksAndHearts.UnitTests/ControllerTests/Landlord/LandlordControllerTests.cs
```diff
@@ -87,4 +87,59 @@ public void EditProfilePage_CalledUsingUserEmail_ReturnsEditProfileViewWithLandl
...
+            .Returns(ILandlordService.LinkUserWithLandlordResult.ErrorLinkDoesNotExist);
+        MakeUserPrincipalInController(landlordUser, UnderTest);
+
+        // Act
+        var result = await UnderTest.TieUserWithLandlord(inviteLink);
+        
+        // Assert
+        result.Should().BeOfType<RedirectToActionResult>().Which.ActionName.Should().Be("Invite");
+        result.Should().BeOfType<RedirectToActionResult>().Which.RouteValues.Should()
+            .Contain("inviteLink",inviteLink);
```
BenRamchandani, line 107: Would be good to check the flash message here as well. Can hopefully just access `UnderTest.TempData["FlashMessage"]` and check it e.g. `.Contains("Invite Link")`

BenRamchandani (changes requested):

web/Services/LandlordService.cs
Chestnut-lol: right I see - sure thank you!

web/Services/LandlordService.cs
```diff
@@ -120,4 +137,42 @@ public bool CheckForDuplicateEmail(LandlordProfileModel editModel)
         return _dbContext.Landlords.SingleOrDefault(l => l.Email == editModel.Email) != null 
                && editedLandlord.Email != editModel.Email;
     }
+    
+    // Link existing landlord with user
+    public async Task<ILandlordService.LinkUserWithLandlordResult> LinkExistingLandlordWithUser(
+        string inviteLink,
+        BricksAndHeartsUser user)
+    {
+        // We want to atomically update multiple records (insert a landlord, then set the user's landlord id), so first start a transaction
```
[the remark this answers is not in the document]
Chestnut-lol: Yeah I would love to learn more about it!

BricksAndHearts.UnitTests/ControllerTests/Landlord/LandlordControllerTests.cs
Chestnut-lol: yeah sure done

BenRamchandani: Don't need the `?` in `UnderTest.TempData["FlashMessage"].Should().BeNull()` or in the others below. In fact you shouldn't have them because then if it were null the assertion will not run.

`Should` here is an extension method and can accept null.

Chestnut-lol: sure done :-)

BenRamchandani (approved):
Looks good!

BenRamchandani (approved):

BenRamchandani (approved):

PR #69, opened 2022-08-01
BNH-69 Enter button

TomHanmer29:
User can now press enter anywhere to move pages forwards and the js error has been fixed from ticket 43


web/wwwroot/js/site.js
```diff
@@ -1,5 +1,25 @@
 ﻿let currentTab = 0;
-showTab(currentTab);
+
+// Get the input field
+let inputs, currentInput
+if (window.location.pathname === "/landlord/register" || window.location.pathname === "/landlord/edit") {
```
MaxTheComputerer, line 5: I'm wondering if it might be nicer to have this as a function which is called from within a `<script>` tag on each of these pages?

web/wwwroot/js/site.js
```diff
@@ -1,5 +1,25 @@
...
+    inputs = document.getElementsByTagName('input');
+    // Execute a function when the user presses a key on the keyboard
+    window.addEventListener("keyup", function (event) {
+        // If the user presses the "Enter" key on the keyboard
+        if (event.key === "Enter") {
+            // Cancel the default action, if needed
+            event.preventDefault();
+            // Deselects all boxes
+            for (currentInput = 0; currentInput < inputs.length; currentInput++) {
+                inputs[currentInput].blur();
```
MaxTheComputerer, line 15: Is there a way to also deselect the next/previous buttons? I've noticed they stay highlighted after they're clicked.

web/wwwroot/js/site.js
TomHanmer29: I did try this at one point a few versions ago and it did seem to break it, however it might work now that I've refactored some of the code. I can give it another go :)

web/wwwroot/js/site.js
TomHanmer29: Yeah this should be pretty straightforward to put in!

web/wwwroot/js/site.js
TomHanmer29: Yeah unfortunately if i ever try and call a function from the script tag in one of these pages it always tells me the function is not defined...

BenRamchandani (approved):
I haven't tested it, but this looks good to me.

Trying to replicate browser behaviour in javascript is always difficult

PR #70, opened 2022-08-01
Bnh 41: delete property

blancamc00:
image
image
Properties can be deleted from the individual view page. Confirmation is required so that they are not deleted accidentally.

### DIFF
--- a/web/Controllers/PropertyController.cs
+++ b/web/Controllers/PropertyController.cs
@@ -131,14 +131,22 @@ public class PropertyController : AbstractController
         return RedirectToAction("ViewProperties", "Landlord");
     }
     
+    [HttpPost]
     [Authorize(Roles = "Landlord")]
+    [Authorize(Roles = "Admin")]
     public ActionResult DeleteProperty(int propertyId)
     {
         PropertyDbModel? propDB = _propertyService.GetPropertyByPropertyId(propertyId);
         if (propDB == null)
         {
             _logger.LogWarning("Property with ID {PropertyId} does not exist", propertyId);
-            return RedirectToAction("Error", "Home", new { status = 404 });
+            return StatusCode(404);
+        }
+
+        if (GetCurrentUser().IsAdmin == false & GetCurrentUser().LandlordId != propDB.LandlordId)
+        {
+            _logger.LogWarning("Property with ID {PropertyId} is not your property.", propertyId);
+            return StatusCode(405);
         }
 
         // Delete property
--- a/web/Controllers/PropertyController.cs
+++ b/web/Controllers/PropertyController.cs
@@ -132,8 +132,7 @@ public class PropertyController : AbstractController
     }
     
     [HttpPost]
-    [Authorize(Roles = "Landlord")]
-    [Authorize(Roles = "Admin")]
+    [Authorize(Roles = "Landlord, Admin")]
     public ActionResult DeleteProperty(int propertyId)
     {
         PropertyDbModel? propDB = _propertyService.GetPropertyByPropertyId(propertyId);
@@ -145,8 +144,8 @@ public class PropertyController : AbstractController
 
         if (GetCurrentUser().IsAdmin == false & GetCurrentUser().LandlordId != propDB.LandlordId)
         {
-            _logger.LogWarning("Property with ID {PropertyId} is not your property.", propertyId);
-            return StatusCode(405);
+            _logger.LogWarning("You do not have access to any property with ID {PropertyId}.", propertyId);
+            return StatusCode(404);
         }
 
         // Delete property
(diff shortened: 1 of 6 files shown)


web/Services/PropertyService.cs
```diff
@@ -94,6 +95,11 @@ public void DeleteProperty(PropertyDbModel property)
         return _dbContext.Properties.SingleOrDefault(p => p.LandlordId == landlordId && p.IsIncomplete == true);
     }
 
+    public PropertyDbModel GetDbModelFromViewModel(PropertyViewModel prop)
```
BenRamchandani, line 98: Don't think this is used

web/Controllers/PropertyController.cs
```diff
@@ -140,6 +140,24 @@ public ActionResult AddNewProperty_Cancel()
...
     }
+    
+    [Authorize(Roles = "Landlord")]
+    public ActionResult DeleteProperty(int propertyId)
+    {
+        PropertyDbModel? propDB = _propertyService.GetPropertyByPropertyId(propertyId);
+        if (propDB == null)
+        {
+            _logger.LogWarning("Property with ID {PropertyId} does not exist", propertyId);
+            return RedirectToAction("Error", "Home", new { status = 404 });
```
BenRamchandani, line 151: This should do almost the same thing (it will not redirect, instead just returning a 404 straight away, but that's probably better)
```suggestion
            return StatusCode(404);
```

web/Controllers/PropertyController.cs
```diff
@@ -140,6 +140,24 @@ public ActionResult AddNewProperty_Cancel()
         // Go to View Properties page
         return RedirectToAction("ViewProperties", "Landlord");
     }
+    
+    [Authorize(Roles = "Landlord")]
+    public ActionResult DeleteProperty(int propertyId)
+    {
+        PropertyDbModel? propDB = _propertyService.GetPropertyByPropertyId(propertyId);
```
BenRamchandani, line 147: Need to check that it's this landlord's property! `[Authorize(Roles = "Landlord")]` just checks that they are a landlord of some kind. It would be nice if admins can delete any property as well.

web/Controllers/PropertyController.cs
```diff
@@ -140,6 +140,24 @@ public ActionResult AddNewProperty_Cancel()
         // Go to View Properties page
         return RedirectToAction("ViewProperties", "Landlord");
     }
+    
+    [Authorize(Roles = "Landlord")]
+    public ActionResult DeleteProperty(int propertyId)
```
BenRamchandani, line 145: This should be a POST - having GET requests cause side effects can cause issues. It will normally be OK in practice, but the browser is within its rights to prefetch the link, retry the request or cache the response.

You'll need to change the `<input>` to be in a `<form method="POST">` tag in the view

BenRamchandani (changes requested):

web/Controllers/PropertyController.cs
```diff
@@ -140,6 +140,32 @@ public ActionResult AddNewProperty_Cancel()
         // Go to View Properties page
         return RedirectToAction("ViewProperties", "Landlord");
     }
+    
+    [HttpPost]
+    [Authorize(Roles = "Landlord")]
+    [Authorize(Roles = "Admin")]
```
BenRamchandani, line 146: I think stacking Authorize attributes does an AND, so you have to be both a landlord and an admin - I think you instead want `[Authorize(Roles="Landlord,Admin")]`, but do check!

web/Controllers/PropertyController.cs
```diff
@@ -140,6 +140,32 @@ public ActionResult AddNewProperty_Cancel()
...
+        if (propDB == null)
+        {
+            _logger.LogWarning("Property with ID {PropertyId} does not exist", propertyId);
+            return StatusCode(404);
+        }
+
+        if (GetCurrentUser().IsAdmin == false & GetCurrentUser().LandlordId != propDB.LandlordId)
+        {
+            _logger.LogWarning("Property with ID {PropertyId} is not your property.", propertyId);
+            return StatusCode(405);
```
BenRamchandani, line 159: 405 is method not allowed (e.g. you did a POST but this endpoint only accepts GET).

This should either be 403 Forbidden, or 404 Not Found if we don't want other users to even be able to determine whether the property exists or not.

BenRamchandani (changes requested):
Couple more things!

BenRamchandani (dismissed):

web/Services/PropertyService.cs
BenRamchandani: This has reappeared!

BenRamchandani (approved):

PR #71, opened 2022-08-01
Enforce read only context in tests

BenRamchandani:


MaxTheComputerer (approved):
Looks good, great idea!

PR #73, opened 2022-08-01
BNH-52 - upload/view/delete images with azure storage

cec200wire:
There is a page, accessible through the 'Images' button on the individual Property Overview pages, which lists all images currently uploaded for that property with options to view and delete them, as well as an option to upload new images:
image
image

No significant effort has been put into the formatting of the Image page, as I suspect much of this will be absorbed into Property Edit pages (once this exists), and the images will need displaying in various places. This branch was aiming primarily towards back-end functionality rather than front-end niceness.

In order for the connection to Azure to work, the BlobConnectionString and BlobContainerName need adding to the appsettings(.Development).json file - I have updated this file on Keeper. I've been setting the BlobContainerName as 'test-' in the Development settings to provide a prefix which distinguishes our testing from any real images; multiple people accessing the same 'test-' containers shouldn't break anything, it should just mean that everyone accesses any images uploaded for testing purposes.

The actual display of images still requires a lot of work (currently images open full-page, as files, which is obviously not ideal) - I think it would be best to open a separate ticket for this, as image display is a separate issue to the Azure Blob Storage things and this branch is already quite large.


web/Controllers/PropertyController.cs
```diff
@@ -151,7 +139,55 @@ public ActionResult ViewProperty(int propertyId)
...
+    public async Task<IActionResult> AddPropertyImages(List<IFormFile> images, int propertyId)
+    {
+        var user = GetCurrentUser();
+        if (!user.IsAdmin && _propertyService.GetPropertyByPropertyId(propertyId)!.LandlordId != user.LandlordId)
+        {
+            return StatusCode(403);
+        }
+        foreach (var image in images)
+        {
+            if (image.Length > 0)
```
BenRamchandani, line 164: When does this happen (length is zero)?

web/Controllers/PropertyController.cs
```diff
@@ -151,7 +139,55 @@ public ActionResult ViewProperty(int propertyId)
...
+                await _azureStorage.UploadFileAsync(image, "property", propertyId);
+            }
+        }
+        return RedirectToAction("ListPropertyImages", "Property", new{propertyId});
+    }
+    
+    [HttpPost("displayImage")]
+    public async Task<IActionResult> DisplayPropertyImage(int propertyId, string fileName)
+    {
+        var image = await _azureStorage.DownloadFileAsync("property", propertyId, fileName);
```
BenRamchandani, line 175: Should check permissions here too (probably want a helper method!)

web/Services/AzureStorage.cs
```diff
@@ -0,0 +1,124 @@
...
+                var newContainerClient = new BlobContainerClient(_storageConnectionString, "default");
+                blobName = "error.png";
+                blobClient = newContainerClient.GetBlobClient(blobName);
+            }
+            Stream data = await blobClient.OpenReadAsync();
+            string fileType = SplitFileName(blobName).type;
+            return (data, fileType);
+        }
+
+        public async Task DeleteFileAsync(string varType, int id, string blobName)
```
BenRamchandani, line 108: Don't feel like you need to call everything "SomethingAsync" unless you're providing both async and non-asnyc methods (which we don't need to)

web/Services/AzureStorage.cs
```diff
@@ -0,0 +1,124 @@
...
+            BlobContainerClient containerClient = new BlobContainerClient(_storageConnectionString, containerName);
+            if (!containerClient.Exists())
+            {
+                await CreateContainerAsync(varType, id);
+                containerClient = new BlobContainerClient(_storageConnectionString, containerName);
+            }
+            return containerClient;
+        }
+
+        public async Task CreateContainerAsync(string varType, int id)
```
BenRamchandani, line 46: Can be private?

web/Services/AzureStorage.cs
```diff
@@ -0,0 +1,124 @@
...
+        private string GetContainerName(string varType, int id)
+        {
+            return _baseContainerName + varType + id.ToString();
+        }
+
+        private async Task<BlobContainerClient> GetContainerClient(string varType, int id)
+        {
+            string containerName = GetContainerName(varType, id);
+            BlobContainerClient containerClient = new BlobContainerClient(_storageConnectionString, containerName);
+            if (!containerClient.Exists())
```
BenRamchandani, line 38: I wonder if we should have a separate `GetOrCreateContainerClient` method with this logic in, just to be clearer what's going on?

web/Services/AzureStorage.cs
```diff
@@ -0,0 +1,124 @@
...
+            {
+                await blobClient.UploadAsync(data);
+            }
+        }
+
+        public async Task<List<string>> ListFilesAsync(string varType, int id)
+        {
+            var containerClient = await GetContainerClient(varType, id);
+            List<string> fileNames = new List<string>();
+            fileNames.Add(id.ToString());
```
BenRamchandani, line 85: Is this intentional, add the id as a file name?

web/Services/AzureStorage.cs
```diff
@@ -0,0 +1,124 @@
...
+        public async Task<(Stream, string)> DownloadFileAsync(string varType, int id, string blobName)
+        {
+            var containerClient = await GetContainerClient(varType, id);
+            BlobClient blobClient = containerClient.GetBlobClient(blobName);
+            if (!blobClient.Exists())
+            {
+                var newContainerClient = new BlobContainerClient(_storageConnectionString, "default");
+                blobName = "error.png";
+                blobClient = newContainerClient.GetBlobClient(blobName);
+            }
```
BenRamchandani, line 102: I think don't bother with this - just return null and the endpoint can return a 404.

web/appsettings.json
```diff
@@ -2,6 +2,8 @@
   "ApplicationInsights": {
     "ConnectionString": ""
   },
+  "BlobConnectionString" : "DefaultEndpointsProtocol=https;AccountName=bnhpropertyimages;AccountKey=6kjyiLjh7FBBr/G+vCf9mWiGDdbsAuHjCXfj7mrsl06ai5H7uRMhPMHlxqx6JxknsBfqY8pMyKI4+AStAOwVHQ==;EndpointSuffix=core.windows.net",
```
BenRamchandani, line 5: leave as empty placeholder

```suggestion
  "BlobConnectionString" : "",
```

web/Services/AzureStorage.cs
cec200wire: Indeed - there was a stage when a container was created as part of property creation, thus it had to be public, but in order to avoid problems with non-existent containers that changed and so the method can now be private.

web/Services/AzureStorage.cs
cec200wire: It was a way to allow the property Id to be accessed on the Image List page (so that new images can be added to the correct property) - there might be a way to do this using the routing, but adding the id to the start of the file list seemed more straightforward.

web/Services/AzureStorage.cs
cec200wire: Having double-checked the code, the only time CreateContainer is actually used now is within GetContainerClient (containers are created the first time they are accessed); I'll combine the two methods into a single GetOrCreateContainerClient, which should be clearer and avoid repeating so many steps.

web/Controllers/PropertyController.cs
cec200wire: Surely the point of uploading images is so that they can be seen by potential tenants, and therefore we shouldn't check permissions here? Hence why I also didn't put a permissions check on ListAllProperties, so that these images would be accessible via that list for potential tenants to view.
There should probably be some permissions settings on the View so that the Add and Delete buttons only appear to the correct users, but I anticipate that 'Display' will end up on some publicly available page while 'Add' and 'Delete' will end up on an Edit page (which has its own permissions checks), hence effectively negating the need for that logic within the view itself.

web/Controllers/PropertyController.cs
cec200wire: I believe the length is zero if the file uploaded (through some sort of glitch) contains no data - it seemed sensible to not allow the upload of an empty/damaged file.

SleepyUser (dismissed):

web/Controllers/PropertyController.cs
BenRamchandani: For now we should be checking permissions and info about a property should only be available to its landlord and admins.

In the long run yes, and I hope we'll get to start implementing that tenants viewing pages - however I think it will still be fairly restricted, with tenants just being able to see properties they've been matched with and linked to rather than browsing the site like RightMove.

web/Services/AzureStorage.cs
BenRamchandani: I think it would be nicer to have a view model class that has the property id and a list of files. That would be easier for someone else reading the code (me included) to quickly understand.

web/Views/Landlord/Properties.cshtml
```diff
@@ -24,7 +24,8 @@
     }
     @foreach (var prop in Model.Properties)
     {
-        <tr>
+        <tr class ="clickable"
+            onclick="location.href = '@Url.Action("ViewProperty", "Property", new { propertyId = prop!.PropertyId })'">
```
BenRamchandani, line 28: Unless it breaks the styling or something use an `<a>` tag around the table row rather than using javascript

SleepyUser: What are the reasons that we avoid Javascript wherever possible? Is it slower to be interpreted by a browser? Or is it more susceptible to errors or something, or for consistency?

web/Controllers/PropertyController.cs
BenRamchandani: Fair enough. Maybe log a warning in this case just so we can see when it happens.

web/Views/Landlord/Properties.cshtml
BenRamchandani: Using the elements built in to the browser over JS gives you all the built in behaviours e.g. you'll get a clicking hand cursor, and the link that would open will appear at the bottom of the page, and middle click to open in new tab will work, screen readers will have an easier time interpreting it.

You can kind of replicate all of those in CSS and JS, but it's more work and easy to get wrong. There's also some people that browse with JS off, though that's pretty uncommon these days.

SleepyUser: That's interesting, thank you! It makes sense

cec200wire: Using the <a> tag currently destroys the styling, so I've left it how it is. If the tables are going to continue having clickable rows, I could raise a ticket to look into this (the Landlord List visible to admins currently uses the same JS, so it's not just this table which is affected)?

BenRamchandani: That's fine, if it breaks the styling don't worry

web/Services/AzureStorage.cs
```diff
@@ -0,0 +1,115 @@
...
+            await foreach (BlobItem file in containerClient.GetBlobsAsync())
+            {
+                fileNames.Add(file.Name);
+            }
+            ImageListViewModel imageList = new ImageListViewModel()
+            {
+                PropertyId = id,
+                FileList = fileNames
+            };
+            return imageList;
```
BenRamchandani, line 83: Don't ned to change it, but as a minor point, usually Controllers will construct ViewModels rather than the service - the idea being the service method can then be reused easily between controllers rather than being tied to a particular page.

web/Services/PropertyService.cs
```diff
@@ -99,4 +101,18 @@ public void DeleteProperty(PropertyDbModel property)
...
+        {
+            return true;
+        }
+        var propertyLandlordId = GetPropertyByPropertyId(propertyId)!.LandlordId;
+        var userLandlordId = currentUser.LandlordId;
+        if (propertyLandlordId == userLandlordId)
+        {
+            return true;
+        }
+        return false;
```
BenRamchandani, line 116: Can shorten to
```suggestion
        return propertyLandlordId == currentUser.LandlordId;;
```
think just as easy to understand

BenRamchandani (approved):

PR #75, opened 2022-08-01
BNH-65 Send emails to a list of admins

Chestnut-lol:


web/Services/MailService.cs
```diff
@@ -31,26 +31,27 @@ public MailService(IOptions<EmailConfigOptions> config)
         string msgToName = ""
```
BenRamchandani, line 31: We should probably make message body subject required arguments, and put a sensible value in for the subject. Can be another PR though

BenRamchandani (approved):
Looks good. Don't forget to update appsettings.Development.json in Keeper when you merge 🙂

web/Services/MailService.cs
Chestnut-lol: Can I add it onto another ticket that I'm working on? (the one that's dealing with sending email exceptions)

BenRamchandani: Sure

PR #77, opened 2022-08-01
BNH-45 Admin edit landlord profile

TomHanmer29:
Admins can now edit landlord profiles by clicking / typing in the route in a similar way to navigating to the profiles. Also added a check that the profile passed into the edit method is properly defined. Wrote additional unit tests.


BenRamchandani (approved):
LGTM

PR #79, opened 2022-08-01
Bnh 84 postcode autocomplete also updates lat lon

Chestnut-lol:


BricksAndHearts.UnitTests/ServiceTests/Api/ApiServiceTests.cs
```diff
@@ -14,23 +15,20 @@
...
 {
     // A fake options
     private static readonly IOptions<AzureMapsOptions> _options = A.Fake<IOptions<AzureMapsOptions>>();
+    private static readonly ILogger<AzureMapsAzureMapsApiService> _logger = A.Fake<ILogger<AzureMapsAzureMapsApiService>>();
+    private static readonly HttpMessageHandler _messageHandler = A.Fake<HttpMessageHandler>();
+    private static readonly HttpClient _httpClient = new HttpClient(_messageHandler);
+    
     // A real API service
-    private readonly AzureMapsAzureMapsApiService _underTest = new(null!, _options);
+    private static readonly AzureMapsAzureMapsApiService _underTest = new(_logger, _options,_httpClient);
```
BenRamchandani, line 27: Avoid static test fakes/services - they will be reused between tests which can lead to weird bugs. Make them in the constructor instead

BricksAndHearts.UnitTests/ServiceTests/Api/ApiServiceTests.cs
```diff
@@ -67,12 +65,12 @@ public void TurnResponseBodyToModel_WhenCalledWithNonEmptyString_ReturnsNonEmpty
...
-            return;
-        }
-        results.Address.Keys.ToList().Should().Contain("streetName");
-        results.Address["streetName"].Should().Be("Adam & Eve Street");
+        results.LatLon.Should().NotBeNull("LatLon should not be null");
+
+        results.Address?.Keys.ToList().Should().Contain("streetName");
+        results.Address?["streetName"].Should().Be("Adam & Eve Street");
+        results.LatLon?["lat"].Should().Be("52.2046");
+        results.LatLon?["lon"].Should().Be("0.13244");
```
BenRamchandani, line 73: Avoid `?` in assertions unless you really only want to run the assertion if the thing is non-null. Here I'm pretty sure we want it to fail if `results.LatLon` is null
```suggestion
        results.LatLon["lat"].Should().Be("52.2046");
        results.LatLon["lon"].Should().Be("0.13244");
```

web/Database/PropertyDbModel.cs
```diff
@@ -23,4 +23,6 @@ public class PropertyDbModel
     public DateTime? CreationTime { get; set; }
     public int? Rent { get; set; }
     public string? Description { get; set; }
+    public string? Lat { get; set; } // latitude
+    public string? Lon { get; set; } // longitude
```
BenRamchandani, line 27: Feels odd to store them as strings. `FLOAT` (C#'s `double` type) should work, or you could use `DECIMAL`.

Azure probably returns them as strings to make sure they keep precision (lots of decimal places), but we don't need it to be super precise.

BenRamchandani (changes requested):

BricksAndHearts.UnitTests/ServiceTests/Api/ApiServiceTests.cs
Chestnut-lol: Yeah the reason I added the question mark is because without it, Rider will complain and I can't pass the github tests.

Chestnut-lol: Because of the possible null reference

BenRamchandani: In that case use `!`.

* `Obj?.Method()` means if `Obj` is null then do nothing
* `Obj!.Method()` means try and call it anyway, and fail with a null pointer exception if `Obj` is null

Generally you shouldn't use `!` much in regular code, but in tests it's fine.

web/Migrations/20220803104341_AddLatLonToProperty.cs
```diff
@@ -0,0 +1,35 @@
...
+namespace BricksAndHearts.Migrations
+{
+    public partial class AddLatLonToProperty : Migration
+    {
+        protected override void Up(MigrationBuilder migrationBuilder)
+        {
+            migrationBuilder.AddColumn<decimal>(
+                name: "Lat",
+                table: "Property",
+                type: "decimal(18,2)",
```
BenRamchandani, line 14: `DECIMAL(18, 2)` means 18 decimal digits of precision, with 16 of them to the left of the decimal point, and 2 to the right - this works well for money, but doesn't make sense for us.

Latitudes and longitudes go from 0 to 360 (or -180 to 180), so we want three digits to the left of the point. Nine decimal places is plenty so `DECIMAL(12, 9)` would work.

BenRamchandani (changes requested):

BenRamchandani (approved):

PR #81, opened 2022-08-02
BNH-66 feat: Make sending emails async and catch exceptions

Chestnut-lol:


web/Controllers/LandlordController.cs
```diff
@@ -58,15 +58,25 @@ public async Task<ActionResult> RegisterPost([FromForm] LandlordProfileModel cre
...
+                try
+                {
+                    await _mailService.SendMsg(msgBody,subject);
+                    _logger.LogInformation("Successfully send email notifications");
+                }
+                catch (Exception e)
+                {
+                    // ignored
+                    _logger.LogWarning("Failed to send email notification. Error: {E}", e);
+                }
```
BenRamchandani, line 79: Could you move this into `MailService`, maybe as a `TrySendMsg` function

web/Controllers/LandlordController.cs
```diff
@@ -58,15 +58,25 @@ public async Task<ActionResult> RegisterPost([FromForm] LandlordProfileModel cre
...
+                              + $"Title: {createModel.Title}\n"
+                              + $"First Name: {createModel.FirstName}" + "\n"
+                              + $"Last Name: {createModel.LastName}" + "\n"
+                              + $"Company Name: {createModel.CompanyName}" + "\n"
+                              + $"Email: {createModel.Email}" + "\n"
+                              + $"Phone: {createModel.Phone}" + "\n";
+                var subject = "Bricks&Hearts - landlord registration notification";
+                try
+                {
+                    await _mailService.SendMsg(msgBody,subject);
```
BenRamchandani, line 72: This will wait for the email to be sent (or fail to send) before continuing with this method.

If we instead do
```cs
Task.Run(_mailService.SendMsg(msgBody,subject));
```
then this method will carry on and respond to the user while the email is still being sent. Don't actually change the code here because then the exception will not be caught - instead wrap the try/catch in another function and `Task.Run` that

BenRamchandani (changes requested):
Happy to explain myself more if the comments don't make sense!

web/Controllers/LandlordController.cs
```diff
@@ -77,19 +78,16 @@ public async Task<ActionResult> RegisterPost([FromForm] LandlordProfileModel cre
...
                               + $"Company Name: {createModel.CompanyName}" + "\n"
                               + $"Email: {createModel.Email}" + "\n"
                               + $"Phone: {createModel.Phone}" + "\n";
-                _mailService.SendMsg(msgBody);
-                if (!createModel.Unassigned)
-                {
-                    return RedirectToAction("MyProfile");
-                }
+                var subject = "Bricks&Hearts - landlord registration notification";
+                Task.Run( () => _mailService.TrySendMsg(msgBody, subject));
```
BenRamchandani: You can disable this warning with a comment, Rider should give you a prompt to do this
```suggestion
#pragma warning disable CS4014
                Task.Run(() => _mailService.TrySendMsg(msgBody, subject));
#pragma warning restore CS4014
```

web/Services/MailService.cs
```diff
@@ -41,15 +48,34 @@ public MailService(IOptions<EmailConfigOptions> config)
...
+        string msgToName = ""
+    )
+    {
+        try
+        {
+            await SendMsg(msgBody,subject);
+        }
+        catch (Exception)
+        {
+            // ignored
```
BenRamchandani, line 78: We should log the exception

BenRamchandani (changes requested):

web/Services/MailService.cs
BenRamchandani: We want to know what the exception was too, so that we can try and work out why it didn't send

Chestnut-lol: ah ok

BenRamchandani (approved):

PR #82, opened 2022-08-02
BNH-78 tests: Added unit tests for all methods in adminService

TomHanmer29:


BricksAndHearts.UnitTests/ServiceTests/Admin/AdminServiceTests.cs
```diff
@@ -72,4 +75,169 @@ public void CancelAdminAccessRequest_SetsHasRequestedAdminToFalse_ForCorrectUser
...
+        var service = new AdminService(context);
+
+        // Act
+        var result = service.GetAdminLists().Result;
+
+        // Assert
+
+        result.Item1.Should().BeOfType<List<UserDbModel>>();
+        result.Item2.Should().BeOfType<List<UserDbModel>>();
+    }
```
BenRamchandani, line 93: This doesn't feel very valuable if we're not checking what's in the lists - the type system should already guarantee those assertions!

BricksAndHearts.UnitTests/ServiceTests/Admin/AdminServiceTests.cs
```diff
@@ -72,4 +75,169 @@ public void CancelAdminAccessRequest_SetsHasRequestedAdminToFalse_ForCorrectUser
...
+    {
+        // Arrange
+        using var context = Fixture.CreateReadContext();
+        var service = new AdminService(context);
+
+        // Act
+        var result = service.GetLandlordDisplayList("").Result;
+        
+        // Assert
+        result.Should().BeOfType<List<LandlordDbModel>>().And.Subject.Should().HaveCount(context.Landlords.ToList().Count);
```
BenRamchandani, line 136:
```suggestion
        result.Should().BeOfType<List<LandlordDbModel>>().And.Subject.Should().HaveCount(context.Landlords.Count());
```

nit don't need the ToList. Doesn't really matter in tests but the `ToList` will load all the landlords into memory as a list, whereas `Count()` should do a `SELECT COUNT(*) FROM` which is quicker

BricksAndHearts.UnitTests/ServiceTests/Admin/AdminServiceTests.cs
```diff
@@ -72,4 +75,169 @@ public void CancelAdminAccessRequest_SetsHasRequestedAdminToFalse_ForCorrectUser
...
+        var service = new AdminService(context);
+        var approvedLandlord = Fixture.CreateApprovedLandlordUser();
+        
+        // Act
+        var result = service.FindUserByLandlordId(1);
+
+        // Assert
+
+        result.Should().BeOfType<UserDbModel>();
+        result!.GoogleAccountId.Should().Be(approvedLandlord.GoogleAccountId);
```
BenRamchandani, line 153: I found this a bit confusing! Makes sense now but worth a comment

```suggestion
        // Arrange
        using var context = Fixture.CreateReadContext();
        var service = new AdminService(context);
        
        // Act
        var result = service.FindUserByLandlordId(1);

        // Assert
        result.Should().NotBeNull();
        // This landlord is created first by the fixture so should have id 1
        var approvedLandlord = Fixture.CreateApprovedLandlordUser()
        result!.GoogleAccountId.Should().Be(approvedLandlord.GoogleAccountId);
```

BricksAndHearts.UnitTests/ServiceTests/Admin/AdminServiceTests.cs
```diff
@@ -72,4 +75,169 @@ public void CancelAdminAccessRequest_SetsHasRequestedAdminToFalse_ForCorrectUser
...
+        // Arrange
+        using var context = Fixture.CreateReadContext();
+        var service = new AdminService(context);
+
+        // Act
+        var result = service.FindUserByLandlordId(1000);
+
+        // Assert
+
+        result.Should().Be(null);
```
BenRamchandani, line 168: Doesn't matter, but

```suggestion
        result.Should().BeNull();
```

BenRamchandani (commented):
Some minor comments, but this looks good

BenRamchandani (approved):

PR #83, opened 2022-08-02
BNH-86 flash message into layout

TomHanmer29:
Moved 2 instances of flashmessage out of individual pages and into layout just once


BenRamchandani (changes requested):
Couple of things:

* Can you continue to use "FlashType" that the admin page uses (though still display the message if it's null)
* Would be good to replace "ApprovalSuccessMessage" on the profile page too

web/Views/Shared/_Layout.cshtml
```diff
@@ -77,6 +77,15 @@
...
 </header>
+@{
+    if (TempData["FlashMessage"] != null)
+    {
+        //uses warning as the default alert colour if there's no value in TempData["FlashType"]
+        <div class="alert alert-@(TempData["FlashType"] ?? "warning")" role="alert">
+                @TempData["FlashMessage"]
+        </div>
+    }
+}
```
MaxTheComputerer, line 88: Could we move this further down somewhere? Maybe in the container or in the `<main>`? I'm not sure I like how it looks when it stretches across the screen 😅

MaxTheComputerer (commented):
All looks good with one question 🙂

web/Views/Shared/_Layout.cshtml
TomHanmer29: Yeah that's a good shout! I'll see if i can put it somewhere slightly less big

MaxTheComputerer (approved):
Looks great, thanks!

BenRamchandani (approved):
LGTM. Next time make separate PRs for unrelated changes!

PR #84, opened 2022-08-02
BNH-81 Landlord type input

TomHanmer29:
Updated landlord type input to have the 4 correct options. Also added a dropdown menu for the landlord to declare if they are for profit/not for profit.
image
image
image
Updated the profile to display these new menu changes


BenRamchandani (approved):
Thanks!

As another PR would you mind renaming `LandlordStatus` to `LandlordType` at some point, unless there's a reason it's like that?

TomHanmer29:
Yes! I noticed this too and it was bugging me for ages, I think past me named it landlordStatus at some point but was never sure. I don't think there's any reason for it, I'll make a new ticket to change it

PR #86, opened 2022-08-02
Bnh 58 admin can add property for landlord

Chestnut-lol:


BenRamchandani (commented):
Thanks for adding comments 🙂 

There's quite a lot of duplication here which I'd like to avoid - it would be better if we can reuse most of the code.

That said I think probably hold off on working on this for now until you've talked to Max and especially Glenn are working on similar things and we don't want to end up with loads of merge conflicts!

Chestnut-lol:
Yeah I'm aware that there might well be a lot of merge conflicts with Glenn's work.

When you say reusing the code, do you mean not having e.g. a method AdminViewProperties, but rather modify ViewProperties to have landlordId as a parameter, rather than getting the landlordId of the current user in the method?

I thought about doing that, but then when we are adding new properties, the property kept being added onto the admin rather than the landlord who we are editing.

I guess there would be a way to get it to work, I was aware that other people might well be working with ViewProperties etc. so I really don't want to make drastic changes to the code, and so I did have a lot of duplications.

Would you prefer me to keep it the way I did it, or I can figure out a way to modify existing functions rather than adding new codes like this?

PR #87, opened 2022-08-03
BNH-82 Add letting profile fields to properties

MaxTheComputerer:
- Letting profile fields have been added to the property form
- Reordered a few of the pages
- Currently unstyled so we can style the whole form together once the design is finalised
- Added some CSS classes for future pill-like radio-checkbox inputs

image


BenRamchandani (approved):
Thanks, looks good

PR #88, opened 2022-08-03
BNH-91 colours

TomHanmer29:
Added basic css variables for colours, styled a couple of buttons


BenRamchandani (changes requested):
Can you check through the buttons `:focus` and `:active` states - some of them are still Bootstrap-primary-blue.

Would be good to make some of the table borders bluey as well, though probably as a separate PR.

BenRamchandani (approved):

BenRamchandani (approved):

PR #89, opened 2022-08-03
BNH 80 - add membershipId

cec200wire:
Adds Membership Id to the landlord registration and edit forms as a field which is displayed and required only if the Charter Status 'Already Approved' is selected (LandlordProvidedCharterStatus==true); otherwise, any input into the MembershipId field is ignored.
MembershipId is displayed in the ChangeAhead Charter section of the profile, if it has been provided; the button to 'Confirm Landlord Charter Membership' for Pending landlords only appears if there is a membership Id. Currently, however, the database will allow landlords without a Membership Id to be in an approved state - this is just not possible by following links through the website.

### DIFF
--- a/web/Views/Shared/_RegisterPagePartial.cshtml
+++ b/web/Views/Shared/_RegisterPagePartial.cshtml
@@ -64,22 +64,6 @@
 </div>
 
 <div class="registerTab" id="ChangeAhead status" >
-   
-    @if (Model.LandlordProvidedCharterStatus)
-    {
-        <style>
-            div.hidetrue {}
-            div.hidefalse {display: none;}
-        </style>
-    }
-    else
-    {
-        <style>
-            div.hidetrue {display: none;}
-            div.hidefalse {}
-        </style>
-    }
-    
     <p>ChangeAhead Charter</p>
 
     <div class="my-3">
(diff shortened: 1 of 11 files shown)


cec200wire:
Profile and Edit/Sign-up with LandlordProvidedCharterStatus==true
image
image

Profile and Edit/Sign-up with LandlordProvidedCharterStatus==false
image
image

web/Views/Shared/_RegisterPagePartial.cshtml
```diff
@@ -1,93 +1,122 @@
...
+        <select asp-for="LandlordProvidedCharterStatus" class="form-select" required
+                id="Select" onclick="checkIfSelectedValueIsTarget()">
+            <option selected disabled value="">Charter Status</option>
+            <option value="true">Already approved</option>
+            <option value="false">Not yet approved</option>
+        </select>
+        <span asp-validation-for="LandlordProvidedCharterStatus" class="text-danger"></span>
+    </div>
+
+    <div id="ifTrue" class="form-group my-3 hidetrue">
```
BenRamchandani, line 96: you could do something like `style="@(Model.LandlordProvidedCharterStatus ? "" : "display: none;")"`, I think that's a bit nicer than having the `<style>` block above

BenRamchandani (approved):
Looks good, thanks 🙂

web/Views/Shared/_RegisterPagePartial.cshtml
```diff
@@ -1,93 +1,122 @@
...
+    
+    <p>ChangeAhead Charter</p>
+
+    <div class="my-3">
+        <label asp-for="LandlordProvidedCharterStatus" class="form-label">What is your ChangeAhead Charter status?</label>
+        <select asp-for="LandlordProvidedCharterStatus" class="form-select" required
+                id="Select" onclick="checkIfSelectedValueIsTarget()">
+            <option selected disabled value="">Charter Status</option>
+            <option value="true">Already approved</option>
+            <option value="false">Not yet approved</option>
```
TomHanmer29, line 75: This looks good! I didn't realise this was a way of doing dropdown lists, seems like a nice solution and I think should help with adding any bootstrap based stuff too.

BenRamchandani: And it lets you do `selected disabled` for a placeholder! That's good, I couldn't get that to work with the `@Html.DropDownList` helper.

BenRamchandani (approved):

web/Views/Shared/_RegisterPagePartial.cshtml
cec200wire: I don't actually think the placeholder works in this case - value="false" overrides it, so the page first loads showing 'Not yet approved' unfortunately.

BenRamchandani: Ah, that might be because `LandlordProvidedCharterStatus` is non-nullable so it defaults to `false`

PR #90, opened 2022-08-03
BNH-82 Letting profile

MaxTheComputerer:
Changes made as a result of testing:
- "Allow pets" radio input is now a required field (you have to choose yes or no)
- Display letting profile on a property's page (looks quite crude, waiting for styling)

### DIFF
--- a/web/Views/Property/ViewProperty.cshtml
+++ b/web/Views/Property/ViewProperty.cshtml
@@ -47,10 +47,39 @@
                     <td> </td>
                 </tr>
             </table>
-            
+
         </div>
         <div class="col-4">
-            
+            <table class="table table-bordered">
+                <tr>
+                    <td>Accepts single tenants</td>
+                    <td>@Model.AcceptsSingleTenant</td>
+                </tr>
+                <tr>
+                    <td>Accepts couples</td>
+                    <td>@Model.AcceptsCouple</td>
+                </tr>
+                <tr>
+                    <td>Accepts families</td>
+                    <td>@Model.AcceptsFamily</td>
+                </tr>
+                <tr>
+                    <td>Allows pets</td>
+                    <td>@Model.AcceptsPets</td>
+                </tr>
+                <tr>
+                    <td>Accepts tenants on Universal Credit/housing benefits</td>
+                    <td>@Model.AcceptsBenefits</td>
+                </tr>
+                <tr>
+                    <td>Accepts tenants not in EET</td>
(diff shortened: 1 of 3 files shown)


web/Views/Property/ViewProperty.cshtml
```diff
@@ -1,83 +1,112 @@
...
+                </tr>
+                <tr>
+                    <td>Accepts tenants not in EET</td>
+                    <td>@Model.AcceptsNotEET</td>
+                </tr>
+                <tr>
+                    <td>Accepts tenants without a rent guarantor</td>
+                    <td>@Model.AcceptsWithoutGuarantor</td>
+                </tr>
+            </table>
```
MaxTheComputerer, line 82: This is the only actual change to this file

web/Views/Property/_AddFormStep5.cshtml
```diff
@@ -25,11 +25,11 @@
...
-        <input type="radio" asp-for="AcceptsPets" value="true" id="PetsTrue" class="form-check-input">
+        <input type="radio" asp-for="AcceptsPets" value="true" id="PetsTrue" class="form-check-input" required>
         <label for="PetsTrue" class="form-check-label">Yes</label>
     </div>
     <div class="form-check">
-        <input type="radio" asp-for="AcceptsPets" value="false" id="PetsFalse" class="form-check-input">
+        <input type="radio" asp-for="AcceptsPets" value="false" id="PetsFalse" class="form-check-input" required>
         <label for="PetsFalse" class="form-check-label">No</label>
     </div>
 </div>
```
cec200wire, line 35: Maybe a similar check should be applied to the fields 'AcceptsSingleTennant', 'AcceptsCouple' and 'AcceptsFamily' - something which requires at least one option (although possibly more than one option, unlike here) to be selected? It's currently possible to leave all of these unselected, which could lead to landlords accidentally providing information (i.e. by misreading the form as 'Not accept' or something) which says they won't accept any tenants.

MaxTheComputerer: Good idea, I'm on it

BenRamchandani: That's a good point, we should definitely implement a check.

Might be easier to do after splitting out the pages to have separate view models though if it's difficult to do now..

MaxTheComputerer: All done :)

BenRamchandani (approved):

PR #92, opened 2022-08-03
BNH-48_fix feat: Names landlord on property list

SleepyUser:
Little fix to name the landlord whose properties are being displayed. 
image
image

Fixed thanks to Alice's feedback on the ticket while testing.

### DIFF
--- a/web/Controllers/LandlordController.cs
+++ b/web/Controllers/LandlordController.cs
@@ -11,9 +11,9 @@ namespace BricksAndHearts.Controllers;
 public class LandlordController : AbstractController
 {
     private readonly ILandlordService _landlordService;
-    private readonly IPropertyService _propertyService;
-    private readonly IMailService _mailService;
     private readonly ILogger<LandlordController> _logger;
+    private readonly IMailService _mailService;
+    private readonly IPropertyService _propertyService;
 
     public LandlordController(ILogger<LandlordController> logger,
         ILandlordService landlordService, IPropertyService propertyService, IMailService mailService)
@@ -77,7 +77,7 @@ public class LandlordController : AbstractController
         {
             case ILandlordService.LandlordRegistrationResult.Success:
                 _logger.LogInformation("Successfully created landlord for user {UserId}", user.Id);
-                var msgBody = $"A Landlord has just registered\n"
+                var msgBody = "A Landlord has just registered\n"
                               + "\n"
                               + $"Title: {createModel.Title}\n"
                               + $"First Name: {createModel.FirstName}" + "\n"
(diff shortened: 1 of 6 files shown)


web/ViewModels/LandlordProfileModel.cs
```diff
@@ -68,4 +68,23 @@ public static LandlordProfileModel FromDbModel(LandlordDbModel landlord, BricksA
...
+            Email = landlord.Email,
+            FirstName = landlord.FirstName,
+            LastName = landlord.LastName,
+            Phone = landlord.Phone,
+            Title = landlord.Title,
+            LandlordType = landlord.LandlordType,
+            LandlordProvidedCharterStatus = landlord.LandlordProvidedCharterStatus,
+            CharterApproved = landlord.CharterApproved,
+            IsLandlordForProfit = landlord.IsLandlordForProfit,
+            CurrentUserIsAdmin = landlord.User?.IsAdmin ?? false
```
BenRamchandani, line 87: We should probably remove `CurrentUserIsAdmin` from LandlordProfileModel and either use `@User.HasRole` or have a separate view model for the page(s) that require it.

If you've got time to do that would you mind making a separate PR for it? If not no worries.

web/Controllers/LandlordController.cs
```diff
@@ -191,7 +191,15 @@ public IActionResult ViewProperties(int? id = null)
 
         var databaseResult = _propertyService.GetPropertiesByLandlord(id.Value);
         var listOfProperties = databaseResult.Select(PropertyViewModel.FromDbModel).ToList();
-        return View("Properties", new PropertiesDashboardViewModel(listOfProperties));
+        var viewModel = new PropertiesDashboardViewModel(listOfProperties);
+        var landlord = await _landlordService.GetLandlordIfExistsFromId(id);
+
+        if (landlord != null)
```
BenRamchandani, line 197: This should never be null right? You could add another method to LandlordService `GetLandlordFromId` that throws an exception if the landlord doesn't exist. Can then make it non-null on the view model.

BenRamchandani (changes requested):
One null check I think we shouldn't do

web/Controllers/LandlordController.cs
SleepyUser: Yes, you're correct! For some reason I was thinking of the situation where an Admin who isn't a landlord tried to visit the /me/ routing. But that would be caught by the nested null check at the start of the function. 

I did consider removing the nullability from the view model and then cast what we had to be non-null to fit, but the method is more reusable.

web/ViewModels/LandlordProfileModel.cs
SleepyUser: I'll do a separate PR for this while Blanca is working on our editing property bits :)

BenRamchandani (approved):

PR #93, opened 2022-08-03
Bnh 94&95 - bug fixes on image upload

cec200wire:
BNH-94: The issue was that when filenames contained spaces, the image wouldn't display. This turned out not to be due to Azure's naming conventions (which are reasonably standard for files) but due to the space being treated as a string delimiter. I have fixed this accordingly (there were some "" missing!).

BNH-95: Added server-side validation of the file type. Currently only JPG, JPEG, PNG, BMP and GIF will be accepted. If there are any other image types which are common enough to be needed, they can easily be added.

I also added some flash messages to indicate successful upload or failed upload due to empty/ non-image file.

### DIFF
--- a/web/Services/AzureStorage.cs
+++ b/web/Services/AzureStorage.cs
@@ -119,7 +119,7 @@ namespace BricksAndHearts.Services
         {
             string fileType = SplitFileName(fileName).type;
             List<string> imageExtensions = new List<string> { "jpg", "jpeg", "png", "bmp", "gif" };
-            if (imageExtensions.Contains(fileType))
+            if (imageExtensions.Contains(fileType.ToLower()))
             {
                 return true;
             }
--- a/web/Services/AzureStorage.cs
+++ b/web/Services/AzureStorage.cs
@@ -70,10 +70,10 @@ namespace BricksAndHearts.Services
 
             if (blob.FileName != fileName)
             {
-                return $"Successfully uploaded {blob.FileName} with name {fileName}";
+                return $"Successfully uploaded {blob.FileName} with name {fileName}.";
             }
 
-            return $"Successfully uploaded {blob.FileName}";
+            return $"Successfully uploaded {blob.FileName}.";
         }
 
         public async Task<ImageListViewModel> ListFiles(string varType, int id)
--- a/web/Services/AzureStorage.cs
+++ b/web/Services/AzureStorage.cs
@@ -7,10 +7,11 @@ namespace BricksAndHearts.Services
     public interface IAzureStorage
     {
         Task DeleteContainer(string varType, int id);
-        Task UploadFile(IFormFile file, string varType, int id);
+        Task<string> UploadFile(IFormFile file, string varType, int id);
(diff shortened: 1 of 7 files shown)


cec200wire:
If multiple files are uploaded at once, only the last flash message is currently displayed. I found a messy way of mostly fixing it, by combining the flash messages into a single message; there might be a better way around this issue, or it might be that it's such a minor issue it's not worth worrying about at the moment. 
`string logInfo = "", flashType = "", flashMessage = "";
int imageCount = 0, flashTypeScale = 0;
foreach (var image in images)
{
    imageCount += 1;
    if (!_azureStorage.IsImage(image.FileName))
    {
        logInfo += $"Failed to upload {image.FileName}: not in a recognised image format";
        flashTypeScale += 1;
        flashMessage +=
            $"{image.FileName} is not in a recognised image format. Please submit your images in one of the following formats: JPG, JPEG, PNG, BMP, GIF";
    }
    else
    {
        if (image.Length > 0)
        {
            string message = await _azureStorage.UploadFile(image, "property", propertyId);
            logInfo += $"Successfully uploaded {image.FileName}";
            flashTypeScale += 0;
            flashMessage += message;
        }
        else
        {
            logInfo += $"Failed to upload {image.FileName}: has length zero.";
            flashTypeScale += 1;
            flashMessage += $"{image.FileName} contains no data, and so has not been uploaded <br>";
        }
    }
}

if (flashTypeScale == 0)
{
    flashType = "success";
}
else if (flashTypeScale == imageCount)
{
    flashType = "danger";
}
else
{
    flashType = "warning";
}
FlashMessage(_logger, (logInfo, flashType, flashMessage));
return RedirectToAction("ListPropertyImages", "Property", new{propertyId});
`

web/Services/AzureStorage.cs
```diff
@@ -105,6 +114,17 @@ public async Task DeleteFile(string varType, int id, string blobName)
                 await blobClient.DeleteAsync();
             }
         }
+        
+        public bool IsImage(string fileName)
+        {
+            string fileType = SplitFileName(fileName).type;
+            List<string> imageExtensions = new List<string> { "jpg", "jpeg", "png", "bmp", "gif" };
+            if (imageExtensions.Contains(fileType))
```
BenRamchandani, line 122: Probably worth lowercasing `fileType` for the comparison

BenRamchandani:
> If multiple files are uploaded at once, only the last flash message is currently displayed. I found a messy way of mostly fixing it, by combining the flash messages into a single message; I've attached a copy of the code (copying it into here really wasn't working for some reason), which would replace lines 194-217 of the PropertyController, if you want to take a look. There might be a better way around this issue, or it might be that it's such a minor issue it's not worth worrying about at the moment.
> 
> [combine flash messages into single flash message.txt](https://github.com/Change-Ahead/bricks-and-hearts/files/9253410/combine.flash.messages.into.single.flash.message.txt)

I think that's worth doing. Maybe merge this and then do it as a separate PR and we'll see if we can get it working

BenRamchandani (approved):

PR #94, opened 2022-08-04
BNH-88 Update navbar and footer

MaxTheComputerer:
- Added secondary navbar for landlords/admins
- Styled navbars to match designs
- Moved privacy and contact links to footer

image

image

image

### DIFF
--- a/web/Views/Shared/_Layout.cshtml
+++ b/web/Views/Shared/_Layout.cshtml
@@ -1,6 +1,7 @@
 ﻿@using BricksAndHearts.Auth
 <!DOCTYPE html>
 <html lang="en">
+
 <head>
     <meta charset="utf-8"/>
     <meta name="viewport" content="width=device-width, initial-scale=1.0"/>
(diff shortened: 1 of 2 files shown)


web/Views/Shared/_Layout.cshtml
```diff
@@ -1,102 +1,104 @@
...
-                        <li class="nav-item">
-                            <a class="nav-link">@(((BricksAndHeartsUser)User.Identity).GoogleEmail)</a>
-                        </li>
-                        <li class="nav-item">
-                            @using (Html.BeginForm("Logout", "Login", FormMethod.Post))
-                            {
-                                <button type="submit" class="btn btn-link nav-link border-0">Log out</button>
-                            }
-                        </li>
+                        @if (User.IsInRole("Landlord"))
```
TomHanmer29, line 36: Am I being silly or does this mean admins won't have a profile picture?

MaxTheComputerer: That's a good point. I think I left it off at the minute because there's no admin profile page, and because this icon is very temporary anyway. I can add one in for admins if you think it's useful?

TomHanmer29: I think its fine for the moment, can always make another ticket if it really feels necessary

web/Views/Shared/_Layout.cshtml
```diff
@@ -1,102 +1,104 @@
...
-                            <a class="nav-link text-dark" asp-area="" asp-controller="Landlord" asp-action="ViewProperties">View Properties</a>
-                        </li>
-                    }
-                </ul>
 
-                <ul class="navbar-nav ms-auto">
+            <div class="collapse navbar-collapse">
+                <div class="navbar-nav me-auto">
+                    @* Change this link to landlord index when created *@
+                    <a class="nav-link text-dark mx-3" asp-controller="Landlord" asp-action="RegisterGet">List your properties</a>
```
BenRamchandani, line 29: Can you do this and just send them to the profile for now?

web/Views/Shared/_Layout.cshtml
```diff
@@ -1,102 +1,104 @@
...
-                        </li>
-                        <li class="nav-item">
-                            @using (Html.BeginForm("Logout", "Login", FormMethod.Post))
-                            {
-                                <button type="submit" class="btn btn-link nav-link border-0">Log out</button>
-                            }
-                        </li>
+                        @if (User.IsInRole("Landlord"))
+                        {
+                            @* Replace with profile picture in the future *@
```
BenRamchandani, line 38: Add "TODO" and the ticket number so we don't miss it
```suggestion
                            @* TODO BNH-99: Replace with profile picture in the future *@

```

web/Views/Shared/_Layout.cshtml
```diff
@@ -1,102 +1,104 @@
...
+                <a asp-controller="Admin" asp-action="GetAdminList" class="nav-link text-light mx-3">Admin list</a>
+                <div class="vr my-2"></div>
+                <a asp-controller="Admin" asp-action="LandlordList" class="nav-link text-light mx-3">Landlord list</a>
+            </nav>
+        }
+        @if (User.IsInRole("Landlord"))
+        {
+            <nav class="navbar navbar-secondary navbar-expand-sm navbar-dark bg-primary py-1">
+                @* Change this link to landlord index when created *@
+                <a asp-controller="Landlord" asp-action="RegisterGet" class="nav-link text-light mx-3">Dashboard</a>
```
BenRamchandani, line 71: (same thing)

BenRamchandani (changes requested):
One random thing: looking at it feels like the logout button should be blue (still outline). What do you think? Don't change it now, we can talk to Shiv about it tomorrow.

web/Views/Shared/_Layout.cshtml
MaxTheComputerer: Do you mean to link to the profile instead of the registration form? (The form redirects to the profile if you're a landlord already). Or to add an "index" action that redirects to the profile?

MaxTheComputerer:
> One random thing: looking at it feels like the logout button should be blue (still outline). What do you think? Don't change it now, we can talk to Shiv about it tomorrow.

Yeah I agree that would look nicer! I was just going off the designs for now

web/Views/Shared/_Layout.cshtml
BenRamchandani: Link to the profile instead of the registration

MaxTheComputerer: Done!

BenRamchandani: Ah sorry I meant if they're registered, link straight to profile, if not link to register. Trying to rush through reviews 😅 

I know register will redirect, but it feels wrong to be relying on that. Maybe that's silly though. Definitely don't want to link straight to profile if they're not registered as it will just error right?

MaxTheComputerer: Yeah profile gives a 403 if they're not registered. I've added your suggestion!

TomHanmer29:
I also think it would be nice to have some sort of indication as to who you actually are on the navbar, from the current one I cant see who i'm signed in as unless I go to my profile which is a bit inconvenient!

MaxTheComputerer:
> I also think it would be nice to have some sort of indication as to who you actually are on the navbar, from the current one I cant see who i'm signed in as unless I go to my profile which is a bit inconvenient!

Yeah that would be good. I guess we'll have the profile pictures eventually, but we could have that be a dropdown which shows your name and a link to your profile? Probably best in a separate ticket though, or worth talking to Shiv about.

BenRamchandani (approved):

BenRamchandani (approved):

PR #100, opened 2022-08-04
Bnh 90 display images

Chestnut-lol:
There is a test that is not working since it involves url.action()

### DIFF
--- a/BricksAndHearts.UnitTests/ControllerTests/Property/PropertyControllerTests.cs
+++ b/BricksAndHearts.UnitTests/ControllerTests/Property/PropertyControllerTests.cs
@@ -310,28 +310,34 @@ public class PropertyControllerTests : PropertyControllerTestsBase
         result!.ViewData.Model.Should().BeOfType<PropertyViewModel>();
     }
     
-    [Fact]
+    // This is not working
+    /*[Fact]
     public async void ListPropertyImages_CallsListFilesAsync_AndReturnsViewListPropertyImages()
     {
         // Arrange
         var adminUser = CreateAdminUser();
         MakeUserPrincipalInController(adminUser, UnderTest);
         A.CallTo(() => PropertyService.IsUserAdminOrCorrectLandlord(adminUser, 1)).Returns(true);
-
-        var model = new ImageListViewModel
-        {
-            PropertyId = 1,
-            FileList = new List<string> { "image1", "image2" }
-        };
-        A.CallTo(() => AzureStorage.ListFiles("property", 1)).Returns(model);
+        
+        var fakeImage = CreateExampleImage();
+        var image = (fakeImage.OpenReadStream(), "jpeg");
+        A.CallTo(() => AzureStorage.DownloadFile("property", 1, fakeImage.FileName)).Returns(image);
+        
+        var urlHelper = A.Fake<IUrlHelper>();
+        // ReSharper disable once Mvc.ActionNotResolved
+        A.CallTo(() => urlHelper.Action("GetImage", new { propertyId = 1, fileName = "TestFileName" })!).Returns("/fake/url");
+        UnderTest.Url = urlHelper;
+        
+        var fileNames = new List<string> { fakeImage.FileName };
+        A.CallTo(() => AzureStorage.ListFileNames("property", 1)).Returns(fileNames);
 
         // Act
         var result = await UnderTest.ListPropertyImages(1) as ViewResult;
 
         // Assert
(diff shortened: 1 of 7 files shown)


Chestnut-lol:
Changed slightly the button on View Property Page:
image

Chestnut-lol:
The images now look like this:
image

BricksAndHearts.UnitTests/ControllerTests/Property/PropertyControllerTests.cs
```diff
@@ -310,28 +310,29 @@ public void ViewProperty_WithExistingProperty_ReturnViewPropertyView()
         result!.ViewData.Model.Should().BeOfType<PropertyViewModel>();
     }
     
+    /*// This is not working
```
BenRamchandani, line 313: Something like this seems to work OK:

```cs
        var urlHelper = A.Fake<IUrlHelper>();
        // ReSharper disable once Mvc.ActionNotResolved
        A.CallTo(() => urlHelper.Action("GetImage", new { propertyId = 1, fileName = "TestFileName" })!).Returns("/fake/url");
        UnderTest.Url = urlHelper;
```

That said I'd like to get this merged, so feel free to just delete the test for now if it's annoying.

web/ViewModels/ImageListViewModel.cs
```diff
@@ -3,5 +3,17 @@
 public class ImageListViewModel
 {
     public int PropertyId { get; set; }
-    public List<string>? FileList { get; set; }
+    public List<ImageFile>? FileList { get; set; }
+}
+
+public class ImageFile
```
BenRamchandani, line 9: Can you rename this, maybe something like `ImageFileUrlModel`? When I see `ImageFile` I assumed it would include the image itself!

web/Controllers/PropertyController.cs
```diff
@@ -179,9 +177,30 @@ public async Task<IActionResult> ListPropertyImages(int propertyId)
...
+            FileList = imageFiles
+        };
+        
         return View(imageList);
     }
+
+    public List<ImageFile> GetFilesFromFileNames(List<string> fileNames, int propertyId)
+    {
+        var imageFiles = new List<ImageFile>();
+        foreach (var fileName in fileNames)
```
BenRamchandani, line 195: You can make this slightly nicer with `fileNames.Select(fileName => {/* Mapping code here*/})` rather than `foreach`

web/Views/Property/ListPropertyImages.cshtml
```diff
@@ -6,27 +6,15 @@
...
-                Delete @file
-                </button>
-            </td>
-        </tr>
-    }
-</table>
+@foreach (var file in Model.FileList!)
+{
+    <figure class="figure">
+        <img src=@file.FileUrl class="figure-img rounded" alt="No image to show :(">
```
BenRamchandani, line 12: `alt` is supposed to be "alternate text" for when the image doesn't load or for people using screen readers, so it should describe the image. So could be `"Property image @file.FileName"`.

BricksAndHearts.UnitTests/ControllerTests/Property/PropertyControllerTests.cs
Chestnut-lol: No I tried but it's still not working... I've commented the code out and perhaps raise a ticket for that?

BenRamchandani: Yeah sure :+1:

web/Controllers/PropertyController.cs
```diff
@@ -179,9 +176,24 @@ public async Task<IActionResult> ListPropertyImages(int propertyId)
...
+        };
+        
         return View(imageList);
     }
+
+    public List<ImageFileUrlModel> GetFilesFromFileNames(List<string> fileNames, int propertyId)
+    {
+        var imageFiles = fileNames.Select(fileName => new ImageFileUrlModel(fileName, Url.Action("GetImage", new { propertyId, fileName })!)).ToList();
+
+        return imageFiles;
```
BenRamchandani, line 195: I think keep the url variable 😄 

```suggestion
        return fileNames.Select(fileName =>
            {
                var url = Url.Action("GetImage", new { propertyId, fileName })!;
                return new ImageFileUrlModel(fileName, url);
            })
            .ToList();
```

BenRamchandani (changes requested):

BenRamchandani (approved):

BenRamchandani (approved):

PR #102, opened 2022-08-04
BNH-108 feat: display multiple flash messages

cec200wire:
Multiple flash messages can now appear on the Image List page - one per file when multiple files are uploaded simultaneously.
image

### DIFF
--- a/web/Controllers/AbstractController.cs
+++ b/web/Controllers/AbstractController.cs
@@ -30,4 +30,18 @@ public abstract class AbstractController : Controller
         TempData["FlashType"] = flash.flashtype;
         TempData["FlashMessage"] = flash.flashmessage;
     }
+    
+    protected void FlashMultipleMessages(ILogger logger, (List<string> logInfo, List<string> flashtypes, List<string> flashmessages) flashes)
+    {
+        var flashTypes = new List<string>();
+        var flashMessages = new List<string>();
+        for (int i = 0; i < flashes.logInfo.Count; i++)
+        {
+            logger.LogInformation(flashes.logInfo[i]);
+            flashTypes.Add(flashes.flashtypes[i]);
+            flashMessages.Add(flashes.flashmessages[i]);
+        }
+        TempData["MultipleFlashTypes"] = flashTypes;
+        TempData["MultipleFlashMessages"] = flashMessages;
+    }
 }
--- a/web/Controllers/AbstractController.cs
+++ b/web/Controllers/AbstractController.cs
@@ -31,16 +31,8 @@ public abstract class AbstractController : Controller
         TempData["FlashMessage"] = flash.flashmessage;
     }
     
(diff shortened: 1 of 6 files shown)


web/Controllers/AbstractController.cs
```diff
@@ -30,4 +30,18 @@ protected void FlashMessage(ILogger logger, (string logInfo, string flashtype, s
...
+        var flashTypes = new List<string>();
+        var flashMessages = new List<string>();
+        for (int i = 0; i < flashes.logInfo.Count; i++)
+        {
+            logger.LogInformation(flashes.logInfo[i]);
+            flashTypes.Add(flashes.flashtypes[i]);
+            flashMessages.Add(flashes.flashmessages[i]);
+        }
+        TempData["MultipleFlashTypes"] = flashTypes;
+        TempData["MultipleFlashMessages"] = flashMessages;
```
MaxTheComputerer, line 45: Would it be possible to just do `TempData["MultipleFlashTypes"] = flashes.flashtypes;` (and the same for `flashmessages`)? Or if you do need to copy it then maybe `TempData["MultipleFlashTypes"] = new List<string>(flashes.flashtypes)`.

web/Controllers/AbstractController.cs
```diff
@@ -30,4 +30,18 @@ protected void FlashMessage(ILogger logger, (string logInfo, string flashtype, s
         TempData["FlashType"] = flash.flashtype;
         TempData["FlashMessage"] = flash.flashmessage;
     }
+    
+    protected void FlashMultipleMessages(ILogger logger, (List<string> logInfo, List<string> flashtypes, List<string> flashmessages) flashes)
```
BenRamchandani, line 34: I think take the logger out.

We should log things as they happen to help with debugging, whereas this will run at the end.

web/Services/AzureStorage.cs
```diff
@@ -115,15 +115,21 @@ public async Task DeleteFile(string varType, int id, string blobName)
...
+                return (true, null);
             }
-            return false;
+
+            string imageExtString = "";
+            foreach (string imageExt in imageExtensions)
+            {
+                imageExtString += imageExt + ", ";
+            }
+            return (false, imageExtString.Substring(0, imageExtString.Length - 2));
```
BenRamchandani, line 132: This is `String.Join` <https://docs.microsoft.com/en-us/dotnet/api/system.string.join?view=net-6.0#system-string-join(system-string-system-collections-generic-ienumerable((system-string)))>

BenRamchandani (changes requested):

web/Controllers/AbstractController.cs
cec200wire: Having removed the logger from here, the entire code just got simpler so I can basically do that now (regardless of if would have worked before or not).

BenRamchandani (approved):

PR #104, opened 2022-08-04
BNH-77 feat: Added tests for all current methods in propertyService

SleepyUser:
Encased each method's tests in a region for organisation

### DIFF
--- a/BricksAndHearts.UnitTests/ServiceTests/Property/PropertyServiceTests.cs
+++ b/BricksAndHearts.UnitTests/ServiceTests/Property/PropertyServiceTests.cs
@@ -1,19 +1,135 @@
-﻿using BricksAndHearts.Services;
+﻿using System.Linq;
+using BricksAndHearts.Services;
 using BricksAndHearts.ViewModels;
 using FluentAssertions;
 using Xunit;
 
 namespace BricksAndHearts.UnitTests.ServiceTests.Property;
 
-public class PropertyServiceTests : IClassFixture<TestDatabaseFixture>
+public class PropertyServiceTests : PropertyServiceTestsBase
 {
-    private TestDatabaseFixture Fixture { get; }
-
     public PropertyServiceTests(TestDatabaseFixture fixture)
     {
         Fixture = fixture;
     }
 
+    private TestDatabaseFixture Fixture { get; }
+
+    #region GetPropertiesByLandlord
+
+    [Theory]
+    [InlineData(1)]
+    [InlineData(2)]
+    public async void GetPropertiesByLandlord_OnlyGetsPropertiesFromInputLandlord(int landlordId)
+    {
+        // Arrange
+        await using var context = Fixture.CreateReadContext();
+        var service = new PropertyService(context);
+
+        // Act
+        var propertiesByLandlord = service.GetPropertiesByLandlord(landlordId);
+
+        // Assert
+        propertiesByLandlord.Should().OnlyContain(u => u.LandlordId == landlordId);
+    }
+
+    #endregion
+
+    #region AddNewProperty
+
+    [Theory]
+    [InlineData(1)]
+    [InlineData(2)]
+    public async void AddNewProperty_AddsNewProperty(int landlordId)
+    {
+        // Arrange
+        await using var context = Fixture.CreateWriteContext();
+        var service = new PropertyService(context);
+        var propertiesBeforeCount = context.Properties.Count();
+        var createModel = new PropertyViewModel
+        {
+            Address = new PropertyAddress
+            {
+                AddressLine1 = "AddNewProperty_AddsNewProperty",
+                County = "Cambridgeshire",
+                TownOrCity = "Cambridge",
+                Postcode = "CB1 1DX"
+            }
+        };
+
+        // Act
+        service.AddNewProperty(landlordId, createModel);
+        context.ChangeTracker.Clear();
+
+        // Assert
+        var property = context.Properties.OrderBy(u => u.CreationTime).Last();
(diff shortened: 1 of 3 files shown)


BricksAndHearts.UnitTests/ServiceTests/Property/PropertyServiceTests.cs
```diff
@@ -1,31 +1,255 @@
...
+
+        // Act
+        service.AddNewProperty(landlordId, createModel);
+        context.ChangeTracker.Clear();
+
+        // Assert
+        var property = context.Properties.OrderBy(u => u.CreationTime).Last();
+        property.AddressLine1.Should().Be("AddNewProperty_AddsNewProperty");
+        property.IsIncomplete = true;
+        property.LandlordId = landlordId;
```
BenRamchandani, line 68: Are these supposed to be asserts?

BricksAndHearts.UnitTests/ServiceTests/Property/PropertyServiceTests.cs
```diff
@@ -1,31 +1,255 @@
...
+    #region DeleteProperty
+
+    [Fact]
+    public async void DeleteProperty_DeletesProperty()
+    {
+        // Arrange
+        await using var context = Fixture.CreateWriteContext();
+        var service = new PropertyService(context);
+        var propertiesBeforeCount = context.Properties.Count();
+        var deleteModel = context.Properties.Where(u => u.Id == 1).ToList()[0];
```
BenRamchandani, line 119:
```suggestion
        var deleteModel = context.Properties.Single(u => u.Id == 1);

```

BricksAndHearts.UnitTests/ServiceTests/Property/PropertyServiceTests.cs
```diff
@@ -1,31 +1,255 @@
...
+
+
+        var adminUser = CreateAdminUser();
+
+
+        context.Properties.First(u => u.Id == 1).LandlordId = OwnerId;
+
+        // Act
+        var allowed = service.IsUserAdminOrCorrectLandlord(adminUser, 1);
+        context.ChangeTracker.Clear();
```
BenRamchandani, line 170: This test is a bit confusing. I don't think the `[InlineData]` helps here. We just need to make sure the landlord Id is different from the admin's user id. We can do that by setting the admin user id like you do below, so let's do that.

The `context.ChangeTracker.Clear();` call here also doesn't do anything since our asserts don't use the database.

BricksAndHearts.UnitTests/ServiceTests/Property/PropertyServiceTests.cs
```diff
@@ -1,31 +1,255 @@
...
+        await using var context = Fixture.CreateWriteContext();
+        var service = new PropertyService(context);
+        var landlordUser = CreateLandlordUser();
+        landlordUser.LandlordId = 50;
+
+        context.Properties.First(u => u.Id == 1).LandlordId = 49;
+
+        // Act
+        var allowed = service.IsUserAdminOrCorrectLandlord(landlordUser, 1);
+        context.ChangeTracker.Clear();
```
BenRamchandani, line 209: Don't need the change tracker clear here. If you wanted to be extra careful you could save the property change and clear the change tracker before the `Act` line

BenRamchandani (changes requested):

BricksAndHearts.UnitTests/ServiceTests/Property/PropertyServiceTests.cs
SleepyUser: Yes :(

BricksAndHearts.UnitTests/ServiceTests/Property/PropertyServiceTests.cs
SleepyUser: I think I misread the part of Max's guide about changetrackers 😅

BricksAndHearts.UnitTests/ServiceTests/Property/PropertyServiceTests.cs
```diff
@@ -1,31 +1,265 @@
...
+
+        // Act
+        service.UpdateProperty(1, updateModel, false);
+        context.ChangeTracker.Clear();
+
+        // Assert
+        var property = context.Properties.Single(u => u.Id == 1);
+        property.AddressLine1.Should().Be("UpdateProperty_UpdatesProperty1");
+        property.IsIncomplete = false;
+        property.AcceptsBenefits = false;
```
BenRamchandani, line 104: Same thing about asserts

BenRamchandani (changes requested):
One more missing bit of asserts, other than that looks good

BricksAndHearts.UnitTests/ServiceTests/Property/PropertyServiceTests.cs
```diff
@@ -1,31 +1,265 @@
...
+        var allowed = service.IsUserAdminOrCorrectLandlord(landlordUser, 1);
+        context.ChangeTracker.Clear();
+
+        // Assert
+        allowed.Should().BeFalse();
+    }
+
+    #endregion
+
+    #region GetPropertyByPropertyID
```
TomHanmer29, line 229: This is nice :) Makes the code very readable!

BenRamchandani (dismissed):

SleepyUser:
Conflict issues mainly surrounding countProperties, will resub asap

BenRamchandani (approved):

PR #105, opened 2022-08-04
BNH-111 fix: move TrySendMsg into MailService

cec200wire:
Hopefully this will have fixed the issue with the MailService causing test failures.

### DIFF
--- a/web/Services/MailService.cs
+++ b/web/Services/MailService.cs
@@ -1,6 +1,4 @@
-﻿using System.Net.Mail;
-using BricksAndHearts.Auth;
-using MailKit.Net.Smtp;
+﻿using BricksAndHearts.Auth;
 using Microsoft.Extensions.Options;
 using MimeKit;
 using SmtpClient = MailKit.Net.Smtp.SmtpClient;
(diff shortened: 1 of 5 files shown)


web/Services/MailService.cs
```diff
@@ -1,59 +1,89 @@
...
+        string subject,
+        string msgFromName = "",
+        string msgToName = ""
+    )
+    {
+        var loggerMessage = "";
+        try
+        {
+            await SendMsg(msgBody, subject, msgFromName, msgToName);
+            loggerMessage = "Successfully sent emails";
```
BenRamchandani, line 78: I think just log here, you'll need to add a logger into this service.

BenRamchandani (commented):
Background:

When you `await` something you're setting off an asynchronous job, then waiting for it to finish.  

You can also choose not to await it, which means the other job will run in the background while the current function carries on - that's why the test sometimes failed, the background thread sometimes called `TrySendMsg` in time for the test, and sometimes didn't.

---

In this case we don't want to wait for the message to send, so we don't await the `TrySendMsg` call - that's what `Task.Run` was doing (it's slightly safer than just not awaiting something).

Anyway, can you add another message to `MailService`, something like `TrySendMsgInBackground` that sets off `TrySendMsg` with `Task.Run`?

BenRamchandani (approved):

PR #109, opened 2022-08-08
BNH-40 - property availability state

cec200wire:
Availability and AvailableFrom added to property database; this information is displayed in appropriate pages of the website.

### DIFF
--- a/web/Database/PropertyDbModel.cs
+++ b/web/Database/PropertyDbModel.cs
@@ -1,3 +1,4 @@
+using System.ComponentModel.DataAnnotations;
 using System.ComponentModel.DataAnnotations.Schema;
 
 namespace BricksAndHearts.Database;
@@ -8,7 +9,6 @@ public class PropertyDbModel
     // Backend
     public int Id { get; set; }
     public int LandlordId { get; set; }
-    public int? RenterUserId { get; set; } = null;
     public virtual LandlordDbModel Landlord { get; set; } = null!;
     public bool IsIncomplete { get; set; }
     public DateTime? CreationTime { get; set; }
@@ -22,7 +22,7 @@ public class PropertyDbModel
     public string Postcode { get; set; } = string.Empty;
     public decimal? Lat { get; set; } // latitude
     public decimal? Lon { get; set; } // longitude
-    
+
     // Property details
     public string? PropertyType { get; set; }
     public int? NumOfBedrooms { get; set; }
@@ -39,6 +39,11 @@ public class PropertyDbModel
     public bool? AcceptsNotEET { get; set; }
     public bool? AcceptsWithoutGuarantor { get; set; }
 
-    // Rent, deposits, and duration
+    // Rent, deposits, availability and duration
     public int? Rent { get; set; }
+
+    public string Availability { get; set; } = "Draft";
+    [DataType(DataType.Date)]
+    public DateTime? AvailableFrom { get; set; } = null;
+    public int? RenterUserId { get; set; } = null;
 }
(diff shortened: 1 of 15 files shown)


web/Database/PropertyDbModel.cs
```diff
@@ -39,6 +39,11 @@ public class PropertyDbModel
     public bool? AcceptsNotEET { get; set; }
     public bool? AcceptsWithoutGuarantor { get; set; }
 
-    // Rent, deposits, and duration
+    // Rent, deposits, availability and duration
     public int? Rent { get; set; }
+
+    public string Availability { get; set; } = "Draft";
```
BenRamchandani, line 45: It would be nice to have constants for these values. you could add e.g.

```cs
public const string AVAILABILITY_DRAFT = "Draft";
```
to this class and then use that. That prevents typos and makes it easier to change them in the future if we want to.

BenRamchandani (approved):

web/Services/PropertyService.cs
```diff
@@ -157,7 +167,8 @@ public PropertyCountModel CountProperties()
     {
         PropertyCountModel propertyCounts = new PropertyCountModel();
         propertyCounts.RegisteredProperties = _dbContext.Properties.Count();
-        //TODO add counts for properties which are Live, Available (Ticket BNH-40)
+        propertyCounts.LiveProperties = _dbContext.Properties.Count(p => p.Availability != "Draft" && p.Landlord.CharterApproved);
+        propertyCounts.AvailableProperties = _dbContext.Properties.Count(p => p.Availability == "Available");
```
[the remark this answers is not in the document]
cec200wire: Potentially, but I think this is part of a wider problem: currently the availability state is being set mostly arbitrarily, whereas once tenants exist we can impose stricter rules on states such as Occupied; this re-vamp of the availability states should probably also include some functionality that makes 'Available Soon' properties shift to 'Available' once the date is passed (or send a message to the landlord to confirm this?). Basically my point is, I view this branch as a basic creation of these states to allow functionality which relies on them to be made; refining them, and figuring out the best way to do this, will occur later.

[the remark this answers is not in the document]
BenRamchandani: Yep, makes sense. For available/available soon, we could just have "Available", and have the Available From column being non-null imply the "soon" part. But yes, something to iterate on :+1:

BenRamchandani (approved):

web/Database/PropertyDbModel.cs
BenRamchandani: Yes.. there are four ways to do "enums" with our stack:

1. what we're doing, just store a string and hope nobody puts bad values in
2. by default if you make it an Enum EntityFramework will store the int representation (i.e. the first thing in the enum is 1, second is 2). That's fine, but it means reordering the enum in code without manually updating the column means it gets out of sync, and you can't tell what it means from the database alone. It is slightly more efficient, though that's not a big deal here.
3. you can use https://docs.microsoft.com/en-us/dotnet/api/microsoft.entityframeworkcore.storage.valueconversion.enumtostringconverter-1?view=efcore-6.0 to store it as a string instead. Could also optionally add a `CHECK` constraint to the table to make sure there are only those values in it
4. You can have a separate table and FK to it. So have an AvailabilityState table with a number and the enum meaning `(1, "Available")`, and then store the number in the Property table and add a foreign key to AvailabilityState to make sure it can't get out of sync. That's more work, but is otherwise "best" in that it enforces consistency while keeping the database meaningful without the code and allowing you to change enums by updating the AvailabilityState table

3 probably makes the most sense here really if you want to try it in another PR, though Microsoft don't seem to have bothered documenting how to use it, so you'll have to Google.

cec200wire: I don't think I'll bother trying it in a separate PR, what we've already done seems fine to me - I was mostly just curious, since I haven't worked with enums before, and when I've tried to google them things haven't made that much sense. Thanks.

PR #114, opened 2022-08-09
Bnh 117

Chestnut-lol:
Landlord can create public view link for each property:
image
for any user who is not signed in, they may go to: /public/propertyid/{propertyid}/{viewLink} to view the property

If the view link is invalid will show this page:
image
if the view link is valid will show this page:
image


web/Controllers/PropertyController.cs
```diff
@@ -281,4 +282,36 @@ public IActionResult SortProperties(string sortBy, int page = 1, int propPerPage
...
         return View("~/Views/Admin/PropertyList.cshtml", new PropertiesDashboardViewModel(listOfProperties.Skip((page-1)*propPerPage).Take(propPerPage).ToList(),  listOfProperties.Count, null! , page, sortBy));
     }
+    
+    [Authorize(Roles = "Admin")]
+    [Route("/admin/get-public-view-link/{propertyId:int}")]
+    [HttpGet]
+    public ActionResult GetPublicViewLink(int propertyId)
+    {
+        var property = _propertyService.GetPropertyByPropertyId(propertyId);
+        if (property == null) // If landlord already linked to user
```
BenRamchandani, line 292: old comment

web/Controllers/PublicController.cs
```diff
@@ -0,0 +1,90 @@
...
+
+
+
+
+
+
+
+
+
+
```
BenRamchandani, line 90: Lots of whitespace!

BenRamchandani (changes requested):
The logic here looks good.

Can you display the link on the normal property page for admins? Ideally the full link, not just the guid.

web/Controllers/PropertyController.cs
Chestnut-lol: sure

PR #116, opened 2022-08-09
BNH-120 Multiple properties per listing

MaxTheComputerer:
- Added `TotalUnits` and `OccupiedUnits` to Property table to track how many properties are contained in each listing, and how many of these are occupied
- `TotalUnits` defaults to 1, `OccupiedUnits` defaults to 0
- SQL check constraint enforces `OccupiedUnits <= TotalUnits`
- Overall property keeps its availability states, but is forced into the Occupied state if all its units are occupied (i.e. `OccupiedUnits == TotalUnits`)
- Any future work which alters `TotalUnits` or `OccupiedUnits` in a different method should make sure the above holds, but the `UpdateProperty` method will likely be sufficient
- Number of available units is calculated as `TotalUnits - OccupiedUnits`

### DIFF
--- a/web/Services/PropertyService.cs
+++ b/web/Services/PropertyService.cs
@@ -129,11 +129,6 @@ public class PropertyService : IPropertyService
                 // If update succeeds in making the property "available soon", then use its from date
                 dbModel.AvailableFrom = updateModel.AvailableFrom;
             }
-            else
-            {
-                // Keep date the same if we're not updating it here
-                dbModel.AvailableFrom = dbModel.AvailableFrom;
-            }
         }
         else
         {
--- a/web/Services/PropertyService.cs
+++ b/web/Services/PropertyService.cs
@@ -66,7 +66,7 @@ public class PropertyService : IPropertyService
 
             Rent = createModel.Rent,
 
-            Availability = createModel.Availability,
+            Availability = createModel.Availability ?? AvailabilityState.Draft,
             TotalUnits = createModel.TotalUnits ?? 1,
             OccupiedUnits = createModel.OccupiedUnits ?? 0
         };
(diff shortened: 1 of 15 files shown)


web/Services/PropertyService.cs
```diff
@@ -113,13 +112,32 @@ public void UpdateProperty(int propertyId, PropertyViewModel updateModel, bool i
...
+            if (updateModel.Availability == AvailabilityState.AvailableSoon)
+            {
+                // If update succeeds in making the property "available soon", then use its from date
+                dbModel.AvailableFrom = updateModel.AvailableFrom;
+            }
+            else
+            {
+                // Keep date the same if we're not updating it here
+                dbModel.AvailableFrom = dbModel.AvailableFrom;
+            }
```
BenRamchandani, line 136: This doesn't do anything right?

BenRamchandani (dismissed):
Looks good.

What do you think the next steps are to make this discoverable for landlords and easy to use?

web/Services/PropertyService.cs
MaxTheComputerer: Oh yeah, think that's leftover from an old ternary expression

BenRamchandani (approved):

MaxTheComputerer:
> Looks good.
> 
> What do you think the next steps are to make this discoverable for landlords and easy to use?

So I've already added a simple number input to the Add Property form for a landlord to specify how many units their property has, although it might be nicer to move this to the beginning of the form as it appears in the designs.

In terms of altering the number of occupied units, I'd imagine this would be done when an admin creates a match and the landlord shouldn't be able to change this manually?

The availability state and number of total/occupied/available units can be displayed to landlords easily (this is already added to the table).

And then obviously we need to do the CSV upload which will make use of this.

Is there anything else we need to consider?

BenRamchandani:
> So I've already added a simple number input to the Add Property form for a landlord to specify how many units their property has, although it might be nicer to move this to the beginning of the form as it appears in the designs.

I think that would make sense, but let's let Shiv have a think about it

> In terms of altering the number of occupied units, I'd imagine this would be done when an admin creates a match and the landlord shouldn't be able to change this manually?

I think they should be able to - they're probably not only renting to ChangeAhead tenants, so the number might change, or e.g. they might need to remove one to renovate it or something.

PR #117, opened 2022-08-09
BNH-116 - Import tenant data from CSV file

cec200wire:
Unit tests for this to follow shortly - I anticipate there will be things which need changing on this branch, so wanted to open a pull request to allow you an initial look through.

### DIFF
--- a/web/Controllers/AdminController.cs
+++ b/web/Controllers/AdminController.cs
@@ -201,14 +201,29 @@ public class AdminController : AbstractController
     public async Task<ActionResult> ImportTenants(IFormFile csvFile)
     {
         string fileName = csvFile.FileName;
+        if (csvFile.Length == 0)
+        {
+            FlashMessage(_logger, ($"{fileName} has length zero.", "danger",$"{fileName} contains no data. Please upload a file containing the tenant data you would like to import."));
+            return RedirectToAction(nameof(TenantList));
+        }
         if (fileName.Substring(fileName.Length - 3) != "csv")
         {
             FlashMessage(_logger, ($"{fileName} not a CSV file.", "danger",$"{fileName} is not a CSV file. Please upload your data as a CSV file."));
+            return RedirectToAction(nameof(TenantList));
         }
-        else
+        
+        var (columnOrder, flashResponse) = _adminService.CheckIfImportWorks(csvFile);
+        if (flashResponse.FlashTypes.Contains("danger"))
         {
-            await _adminService.ImportTenants(csvFile);
+            FlashMultipleMessages(flashResponse.FlashTypes, flashResponse.FlashMessages);
+            return RedirectToAction(nameof(TenantList));
         }
+        var furtherFlashResponse = await _adminService.ImportTenants(csvFile, columnOrder, flashResponse);
+        if (furtherFlashResponse.FlashMessages.Count() == 0)
+        {
+            FlashMessage(_logger, ("Successfuly imported all tenant data.", "success", "Successfully imported all tenant data."));
+        }
+        FlashMultipleMessages(furtherFlashResponse.FlashTypes, furtherFlashResponse.FlashMessages);
         return RedirectToAction(nameof(TenantList));
     }
 }
--- a/web/Controllers/AdminController.cs
+++ b/web/Controllers/AdminController.cs
@@ -13,16 +13,19 @@ namespace BricksAndHearts.Controllers;
 public class AdminController : AbstractController
 {
     private readonly IAdminService _adminService;
+    private readonly ICsvImportService _csvImportService;
     private readonly ILandlordService _landlordService;
-    private readonly IPropertyService _propertyService;
     private readonly ILogger<AdminController> _logger;
+    private readonly IPropertyService _propertyService;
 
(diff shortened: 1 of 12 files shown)


cec200wire:
Regarding the fact that this is failing the Build and run tests - the warning is related to the LINQtoCSV package I added in order to facilitate the CSV upload; I managed to suppress this warning on local by adding
`<PackageReference Include="LINQtoCSV" Version="1.5.0">
            <NoWarn>NU1701</NoWarn>
        </PackageReference>`
to the csproj file, but I've no idea how to make GitHub accept it. Any ideas?

web/Controllers/AdminController.cs
```diff
@@ -14,25 +14,33 @@ namespace BricksAndHearts.Controllers;
...
     {
         var isAuthenticated = User.Identity?.IsAuthenticated ?? false;
-        var viewModel = new AdminDashboardViewModel(_landlordService.CountLandlords(), _propertyService.CountProperties());
-        if (isAuthenticated) viewModel.CurrentUser = GetCurrentUser();
+        var viewModel =
+            new AdminDashboardViewModel(_landlordService.CountLandlords(), _propertyService.CountProperties());
+        if (isAuthenticated)
+        {
+            viewModel.CurrentUser = GetCurrentUser();
+        }
```
TomHanmer29, line 42: I personally prefer the older version without the brackets but I'm not sure which (if any) of these are "better". I guess both have upsides and downsides :)

cec200wire: I have absolutely no memory of changing that, I think maybe Rider just automatically reformatted it? That is to say, I don't have any strong feelings about this, so could easily change it back if you prefer.

TomHanmer29: Dont worry about it, its probably more effort than its worth :) I just feel like I had nothing else worthwhile to comment on cause the rest of your code looked good!

cec200wire:
Unit tests all added now!

web/Services/CsvImportService.cs
```diff
@@ -0,0 +1,133 @@
...
+
+        //check for extra columns in csv file
+        foreach (var header in headers)
+        {
+            if (!databaseHeaders.Contains(header))
+            {
+                _logger.LogWarning($"Extra column: {header}");
+                flashTypes.Add("warning");
+                flashMessages.Add(
+                    $"The column {header} does not exist in the database. All data in this column has been ignored.");
```
BenRamchandani, line 60: Would be slightly clearer, especially if the problem is trailing whitespace or something.
```suggestion
                    $"The column \"{header}\" does not exist in the database. All data in this column has been ignored.");
```

web/Services/CsvImportService.cs
```diff
@@ -0,0 +1,133 @@
...
+            foreach (var tenant in _dbContext.Tenants)
+            {
+                _dbContext.Tenants.Remove(tenant);
+            }
+            
+            foreach (TenantUploadModel tenant in list)
+            {
+                if (tenant.Name == null)
+                {
+                    _logger.LogWarning($"Name is missing from record {tenant.Email}.");
```
BenRamchandani, line 90: Might be worth logging out the line number too as not all tenants will have emails

web/Views/Admin/TenantList.cshtml
```diff
@@ -38,4 +52,11 @@
         </tr>
     }
     </tbody>
-</table>
\ No newline at end of file
+</table>
+
+@using (Html.BeginForm("ImportTenants", "Admin", FormMethod.Post, new { enctype = "multipart/form-data" }))
+{
+    <div>Please upload a CSV file containing tenant information</div>
```
BenRamchandani, line 59: Can you update this to make it clear that it will replace all existing tenant records?

BenRamchandani (commented):
This looks good, and works well. One thing we should do is provide an example CSV and some instructions, e.g. mentioning that it's whitespace and case sensitive.

I've left a few small comments, but there are also a couple of things we should do as separate PRs/tickets:

I think lets not try and make the import generic by using reflection to get the types of properties, and just keep it simple, i.e.
```cs
dbModel.Name = uploadModel.Name;
dbModel.ETT = parseBool(uploadModel.ETT, flashMessages); // <-- returns true, false or null, and adds a warning to flashMessages if required
etc.
```
I think we'll want to add more validation, e.g. on email and phone, and maybe lookup coordinates based on the postcode, so we'll end up wanting custom code anyway.

I'd also like to tidy up flash messages at some point, we use them a lot! I'm hoping we can get rid of the distinction between multiple and single flash messages and just have one function for it.

BenRamchandani:
> Regarding the fact that this is failing the Build and run tests - the warning is related to the LINQtoCSV package I added in order to facilitate the CSV upload; I managed to suppress this warning on local by adding `<PackageReference Include="LINQtoCSV" Version="1.5.0"> <NoWarn>NU1701</NoWarn> </PackageReference>` to the csproj file, but I've no idea how to make GitHub accept it. Any ideas?

I think you just need to add the same warning suppression in the unit tests `.csproj` file

BenRamchandani (approved):

PR #118, opened 2022-08-09
BNH-114 Charter approval membership ID field

MaxTheComputerer:
Charter status has been simplified to use `membershipId` to test if the landlord has completed the charter, and the `CharterApproved` bool for determining if it's been approved.
Also check membership ID for duplicates when approving

### DIFF
--- a/web/Views/Landlord/_RegisterPagePartial.cshtml
+++ b/web/Views/Landlord/_RegisterPagePartial.cshtml
@@ -71,22 +71,20 @@
         <div class="col-1"></div>
         <div class="col-md-5">
             <div class="form-group my-3">
-                <label asp-for="LandlordProvidedCharterStatus" class="form-label">Charter status</label>
-                <select asp-for="LandlordProvidedCharterStatus" class="form-select" required
+                <label for="Select" class="form-label">Charter status</label>
+                <select class="form-select" required
                         id="Select" onclick="checkIfSelectedValueIsTarget('true')">
-                    <option selected disabled value="">Charter Status</option>
-                    <option value="true">Already approved</option>
-                    <option value="false">Not yet approved</option>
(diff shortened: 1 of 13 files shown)


cec200wire (commented):
Looks good to me, it's definitely a lot simpler without the LandlordHasApprovedCharter property.

web/Views/Landlord/_RegisterPagePartial.cshtml
```diff
@@ -71,22 +71,20 @@
...
-                <label asp-for="LandlordProvidedCharterStatus" class="form-label">Charter status</label>
-                <select asp-for="LandlordProvidedCharterStatus" class="form-select" required
+                <label for="Select" class="form-label">Charter status</label>
+                <select class="form-select" required
                         id="Select" onclick="checkIfSelectedValueIsTarget('true')">
-                    <option selected disabled value="">Charter Status</option>
-                    <option value="true">Already approved</option>
-                    <option value="false">Not yet approved</option>
+                    <!option @(Model.MembershipId != null ? "selected" : "") value="true">Already approved</!option>
+                    <!option @(Model.MembershipId == null ? "selected" : "") value="false">Not yet approved</!option>
```
BenRamchandani, line 78: What do the exclamation marks do?

BenRamchandani (approved):
One question, but otherwise LGTM

web/Views/Landlord/_RegisterPagePartial.cshtml
MaxTheComputerer: They force it to use regular HTML tags instead of the razor tag helpers. I needed this so I could inject the "selected" attribute into the tag, which the tag helpers don't allow you to do!

BenRamchandani: I would expect `<option selected="@(Model.MembershipId == null ? "true" : "false")">` to work, but fine if not

PR #119, opened 2022-08-10
BNH-119 tenant filtering

TomHanmer29:
Created filters for the tenant list.
image
image
The user can select multiple filters which all stack on top of each other to refine the tenant list

### DIFF
--- a/web/Controllers/AdminController.cs
+++ b/web/Controllers/AdminController.cs
@@ -189,13 +189,6 @@ public class AdminController : AbstractController
     }
 
     [Authorize(Roles = "Admin")]
-    [HttpPost]
-    public IActionResult ReceiveTenantFilterList([FromForm] TenantListModel tenantListModel)
-    {
-        return RedirectToAction("GetFilteredTenants", tenantListModel);
-    }
-
-    [Authorize(Roles = "Admin")]
     [HttpGet]
     public async Task<IActionResult> GetFilteredTenants(TenantListModel tenantListModel)
     {
--- a/web/Controllers/AdminController.cs
+++ b/web/Controllers/AdminController.cs
@@ -1,7 +1,6 @@
 #region
 
 using BricksAndHearts.Auth;
-using BricksAndHearts.Database;
 using BricksAndHearts.Services;
 using BricksAndHearts.ViewModels;
 using Microsoft.AspNetCore.Authorization;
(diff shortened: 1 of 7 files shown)


BenRamchandani:
Is there a way to check for e.g. everyone not on universal credit?

TomHanmer29:
> Is there a way to check for e.g. everyone not on universal credit?

Not currently, I will add this though!

web/Controllers/AdminController.cs
```diff
@@ -188,4 +187,19 @@ public ActionResult RejectAdminRequest(int userToAcceptId)
 
         return RedirectToAction("GetAdminList");
     }
+
+    [Authorize(Roles = "Admin")]
+    [HttpPost]
+    public IActionResult ReceiveTenantFilterList([FromForm] TenantListModel tenantListModel)
+    {
+        return RedirectToAction("GetFilteredTenants", tenantListModel);
+    }
```
BenRamchandani, line 196: If you change the form method to GET can you get rid of this endpoint?

web/Services/AdminService.cs
```diff
@@ -147,4 +149,28 @@ public UserDbModel GetUserFromId(int userId)
...
+                tenantQuery = currentFilter switch
+                {
+                    0 => tenantQuery.Where(t => t.Type == "Single"),
+                    1 => tenantQuery.Where(t => t.Type == "Couple"),
+                    2 => tenantQuery.Where(t => t.Type == "Family"),
+                    3 => tenantQuery.Where(t => t.HasPet == true),
+                    4 => tenantQuery.Where(t => t.ETT == true),
+                    5 => tenantQuery.Where(t => t.UniversalCredit == true),
+                    6 => tenantQuery.Where(t => t.HousingBenefits == true),
+                    7 => tenantQuery.Where(t => t.Over35 == true),
```
BenRamchandani, line 169: Could turn this into an object with a nullable bool for each of these options. Slightly less succinct, but a bit simpler to use/understand

BenRamchandani (commented):
> > Is there a way to check for e.g. everyone not on universal credit?
> 
> Not currently, I will add this though!

Sounds good. The code itself looks sensible, though you might have to change it to support yes/no/not filtered

TomHanmer29:
image
image

BenRamchandani (approved):
This all looks good. You can get rid of the old tenant list endpoint if you want and only use the filtered one, just defaulting to an empty filter

TomHanmer29:
I've made a new ticket to improve the styling of these so I can do that as part of a whole refactor then :)

PR #120, opened 2022-08-10
BNH-97 Add address to landlord

MaxTheComputerer:
- Landlords now have an address!
- `PropertyAddress` has been refactored to be `AddressModel` and is shared by properties and landlords
- The address has been added to the landlord register form (mostly at random as this form will be restyled in the future)
- The landlord profile page now displays their address instead of a placeholder, and the corresponding edit button is now functional

### DIFF
--- a/BricksAndHearts.UnitTests/ControllerTests/Landlord/LandLordControllerUnassignedTests.cs
+++ b/BricksAndHearts.UnitTests/ControllerTests/Landlord/LandLordControllerUnassignedTests.cs
@@ -40,11 +40,13 @@ public class LandLordControllerUnassignedTests : LandlordControllerTestsBase
             A<string>.That.Matches(s => s == msgBody), A<string>.That.Matches(s => s == subject), A<string>.Ignored,
             A<string>.Ignored
         )).WithAnyArguments().DoesNothing();
+
         // Act
         var result = await UnderTest.RegisterPost(formResultModel) as RedirectToActionResult;
+
         // Assert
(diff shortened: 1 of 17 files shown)


BenRamchandani (changes requested):
One thing - we should validate the address model server side too, by default `ModelState.IsValid` doesn't check child objects are valid.

BricksAndHearts.UnitTests/ControllerTests/Landlord/LandLordControllerUnassignedTests.cs
```diff
@@ -40,11 +40,16 @@ public async void
             A<string>.That.Matches(s => s == msgBody), A<string>.That.Matches(s => s == subject), A<string>.Ignored,
             A<string>.Ignored
         )).WithAnyArguments().DoesNothing();
+
+        // Fixes NullReferenceException when calling TryValidateModel()
+        UnderTest.ObjectValidator = CustomValidator;
```
BenRamchandani, line 45: Could move this to the test base class constructor, don't see any reason not to have it available for every test.

BenRamchandani (approved):

BricksAndHearts.UnitTests/ControllerTests/Landlord/LandLordControllerUnassignedTests.cs
MaxTheComputerer: That's a good idea - done!

PR #121, opened 2022-08-10
BNH-127 - add AdminControllerTests

cec200wire:
I've added unit tests for all the endpoints in the Admin Controller. I've also done some minor refactoring in the Admin Controller, and to a lesser extent the Admin Service - this was primarily just changing the order in which the endpoints/methods appear, in order to group together ones which are to do with similar issues. I've been finding it increasingly difficult to work out if a relevant method already exists or not, since everything was a bit jumbled up - hopefully having re-ordered things slightly will make this easier for everyone.

### DIFF
--- a/web/Controllers/AdminController.cs
+++ b/web/Controllers/AdminController.cs
@@ -67,14 +67,11 @@ public class AdminController : AbstractController
         if (user.IsAdmin)
         {
             LoggerAlreadyAdminWarning(_logger, user);
-
             return RedirectToAction(nameof(AdminDashboard));
         }
 
         _adminService.CancelAdminAccessRequest(user);
-
         FlashRequestSuccess(_logger, user, "cancelled admin access request");
-
         return RedirectToAction(nameof(AdminDashboard));
     }
     
@@ -90,13 +87,8 @@ public class AdminController : AbstractController
     [HttpPost]
     public ActionResult AcceptAdminRequest(int userToAcceptId)
     {
-        if (_adminService.GetUserFromId(userToAcceptId) == null)
-        {
-            return View("Error", new ErrorViewModel());
-        }
-
         _adminService.ApproveAdminAccessRequest(userToAcceptId);
-
+        _logger.LogInformation($"Admin request of user {userToAcceptId} approved by user {GetCurrentUser().Id}");
         return RedirectToAction("GetAdminList");
     }
 
@@ -104,13 +96,8 @@ public class AdminController : AbstractController
     [HttpPost]
     public ActionResult RejectAdminRequest(int userToRejectId)
     {
-        if (_adminService.GetUserFromId(userToRejectId) == null)
-        {
-            return View("Error", new ErrorViewModel());
-        }
-        
         _adminService.RejectAdminAccessRequest(userToRejectId);
-
+        _logger.LogInformation($"Admin request of user {userToRejectId} rejected by user {GetCurrentUser().Id}");
         return RedirectToAction("GetAdminList");
     }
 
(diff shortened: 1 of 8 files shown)


web/Services/AdminService.cs
```diff
@@ -44,11 +57,43 @@ public void CancelAdminAccessRequest(BricksAndHeartsUser user)
...
 
     public async Task<(List<UserDbModel> CurrentAdmins, List<UserDbModel> PendingAdmins)> GetAdminLists()
     {
         return (await GetCurrentAdmins(), await GetPendingAdmins());
     }
+    
+    private async Task<List<UserDbModel>> GetCurrentAdmins()
+    {
+        var currentAdmins = await _dbContext.Users.Where(u => u.IsAdmin == true).ToListAsync();
+        return currentAdmins;
```
BenRamchandani, line 88: Not your code, but anyway
```suggestion
        return await _dbContext.Users.Where(u => u.IsAdmin == true).ToListAsync();
```

web/Services/AdminService.cs
```diff
@@ -44,11 +57,43 @@ public void CancelAdminAccessRequest(BricksAndHeartsUser user)
...
     }
+    
+    public void ApproveAdminAccessRequest(int userId)
+    {
+        var userToAdmin = _dbContext.Users.Single(u => u.Id == userId);
+        
+        userToAdmin.IsAdmin = true;
+        userToAdmin.HasRequestedAdmin = false;
+        _dbContext.SaveChanges();
+        _logger.LogInformation("Admin request approved");
```
BenRamchandani, line 68: Could you log the user id that we're making an admin as part of this message? or even better move the log to the controller and log both the user we're changing and the current user

web/Controllers/AdminController.cs
```diff
@@ -68,15 +67,49 @@ public IActionResult CancelAdminAccessRequest()
...
+                "Already an admin"));
+    }
+    
+    [Authorize(Roles = "Admin")]
+    [HttpPost]
+    public ActionResult AcceptAdminRequest(int userToAcceptId)
+    {
+        if (_adminService.GetUserFromId(userToAcceptId) == null)
+        {
+            return View("Error", new ErrorViewModel());
```
BenRamchandani, line 85: Not your code again, but tbh I think this should just throw an exception - that will at least log something. I think omit this check completely and let the service method throw an exception (and same for reject)

BenRamchandani (dismissed):
Tests look good, though I've thought of more refactors/changes if you don't mind...

web/Services/AdminService.cs
```diff
@@ -44,11 +55,47 @@ public void CancelAdminAccessRequest(BricksAndHeartsUser user)
         userRecord.HasRequestedAdmin = false;
         _dbContext.SaveChanges();
     }
+    
+    public void ApproveAdminAccessRequest(int userId)
+    {
+        var userToAdmin = _dbContext.Users.SingleOrDefault(u => u.Id == userId);
+        if (userToAdmin == null)
+        {
+            throw new Exception();
```
BenRamchandani, line 64: Worth having some text in there for someone debugging it (same below)
```suggestion
            throw new Exception($"No user found with id {userId}");
```

BenRamchandani (approved):

cec200wire:
I had to rebase again, due to merge conflicts with my own work on another branch.

cec200wire:
Git did something weird and tangled up two of my current branches - this hasn't yet rebased properly. I'll try again.

BenRamchandani (approved):
Still slightly mixed up, the table clickable changes are in here too, but fine by me to just merge anyway

PR #122, opened 2022-08-10
Bnh 113 sort properties by location

Chestnut-lol:


Chestnut-lol:
There is a test missing (I have put in TODO for tomorrow) - but I thought I would first put up a PR anyway since it all seems to be working pretty fine.

image
There is now a textbox for users to enter postcode, and the table will be sorted so that the closest property to the location is at the top of the table

If a property does not have longitude and latitude it will not appear in the table currently

If the postcode entered by the user is not valid, then a warning will be displayed:
image

web/Services/PostcodeApiService.cs
```diff
@@ -0,0 +1,97 @@
...
+        {
+            return new PostcodeioResponseModel();
+        }
+
+        var postcodeResponse =
+            JsonConvert.DeserializeObject<PostcodeioResponseModel>(responseBody);
+        return postcodeResponse!;
+    }
+
+    public async Task<List<PropertyDbModel>?> SortPropertiesByLocation(string postalCode)
```
BenRamchandani, line 62: Minor, but I'd expect this to be a method in property service, that then calls PostcodeApiService to get the location.

BricksAndHearts.UnitTests/ServiceTests/Api/PostcodeApiServiceTests.cs
```diff
@@ -0,0 +1,85 @@
...
+    public PostcodeApiServiceTests()
+    {
+        _logger = A.Fake<ILogger<PostcodeApiService>>();
+        _messageHandler = A.Fake<HttpMessageHandler>();
+        _httpClient = new HttpClient(_messageHandler);
+        _underTest = new PostcodeApiService(_logger, null!, _httpClient);
+    }
+
+    [Fact]
+    public async Task MakeApiRequestToAzureMaps_WhenCalled_ReturnsNonEmptyString()
```
BenRamchandani, line 30: Not Azure

BenRamchandani (commented):
Looks good so far 🙂

web/Services/PostcodeApiService.cs
Chestnut-lol: So I tried doing this but ended up too much codes is being affected especially the testbase. I guess I could make the methods in PostcodeApiService static and access them in PropertyService directly using PostcodeApiService.[method] rather than constructing an instance of PropertyService? Because PropertyService also requires a logger and I'm not too sure as to how to do it properly without breaking anything...

BricksAndHearts.UnitTests/ServiceTests/Property/PropertyServiceTests.cs
```diff
@@ -402,4 +410,74 @@ public void CreateNewPublicViewLink_UpdateDatabase()
         result.Should().NotBeNullOrEmpty();
         context.Properties.Single(p => p.Id == 1).PublicViewLink.Should().Be(result);
     }
+    
+    // TODO: need test for SortPropertiesByLocation
```
BenRamchandani, line 414: This is done 🙂

BricksAndHearts.UnitTests/ServiceTests/Property/PropertyServiceTests.cs
```diff
@@ -402,4 +410,74 @@ public void CreateNewPublicViewLink_UpdateDatabase()
...
+        };
+        var propertyDbModelList = result!.FindAll(p => p.LandlordId == 3);
+        propertyDbModelList.Should().NotBeNullOrEmpty();
+        propertyDbModelList.Should().HaveCount(4);
+        var i = 0;
+        while (i < 4)
+        {
+            propertyDbModelList[i].Should().BeEquivalentTo(correctList[i]);
+            i++;
+        }
```
BenRamchandani, line 481: Can you just do `propertyDbModelList.Should().BeEquivalentTo(correctList);`, or does that not work?

web/Services/PostcodeApiService.cs
```diff
@@ -0,0 +1,62 @@
...
+
+    public PostcodeApiService(ILogger<PostcodeApiService> logger, BricksAndHeartsDbContext dbContext, HttpClient? client = null)
+    {
+        _logger = logger;
+        _dbContext = dbContext;
+        if (client != null)
+        {
+            _client = client;
+        }
+    }
```
BenRamchandani, line 27: Do we need to let it be null? It's fine to still pass in `null!` in the tests if we want to.

```suggestion
    public PostcodeApiService(ILogger<PostcodeApiService> logger, BricksAndHeartsDbContext dbContext, HttpClient client)
    {
        _logger = logger;
        _dbContext = dbContext;
        _client = client;
    }
```

BricksAndHearts.UnitTests/ServiceTests/Property/PropertyServiceTests.cs
Chestnut-lol: I think that the list isn't ordered... as in, it seems to think that [1,2,3,4] is the same as [1,2,4,3] - I might be wrong tho

BenRamchandani (approved):

PR #123, opened 2022-08-11
BNH-123 invite link displays properly and has a copy button

TomHanmer29:
image
image
image

### DIFF
--- a/web/Controllers/AdminController.cs
+++ b/web/Controllers/AdminController.cs
@@ -4,6 +4,7 @@ using BricksAndHearts.Auth;
 using BricksAndHearts.Services;
 using BricksAndHearts.ViewModels;
 using LINQtoCSV;
+using Microsoft.ApplicationInsights.AspNetCore.Extensions;
 using Microsoft.AspNetCore.Authorization;
 using Microsoft.AspNetCore.Mvc;
 
@@ -130,7 +131,7 @@ public class AdminController : AbstractController
                 flashMessageBody = "Landlord already has an invite link";
             }
 
-            FlashMessage(_logger, (flashMessageBody, "success", flashMessageBody + $": {inviteLink}"));
+            FlashMessage(_logger, (flashMessageBody, "success", flashMessageBody + $": {(HttpContext.Request.IsHttps ? "https" : "http")}://{HttpContext.Request.Host}/invite/{inviteLink}"));
         }
 
         return RedirectToAction("Profile", "Landlord", new { id = landlordId });
--- a/web/wwwroot/js/site.js
+++ b/web/wwwroot/js/site.js
@@ -18,8 +18,11 @@ function checkIfSelectedValueIsTarget(target){
 }
 
 function copyLinkToClipboard(linkToCopy){
-    // Copy the text passed into the function
-    console.log(window.location.origin,"/invite/",linkToCopy)
-    console.log(window.location.origin+"/invite/"+linkToCopy)
-    navigator.clipboard.writeText(linkToCopy);
+    // Copy the text passed into the function plus the rest of the url
+    let fullLink = window.location.origin+"/invite/"+linkToCopy
+    navigator.clipboard.writeText(fullLink);
+    
+    // Change the button styling to show the link has been copied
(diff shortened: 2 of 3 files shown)


web/Controllers/AdminController.cs
```diff
@@ -4,6 +4,7 @@
 using BricksAndHearts.Services;
 using BricksAndHearts.ViewModels;
 using LINQtoCSV;
+using Microsoft.ApplicationInsights.AspNetCore.Extensions;
```
BenRamchandani, line 7: Is this import used?

BenRamchandani (approved):
Looks good. Check it works on staging after merging just in case the app service does something weird to the `Host` variable

web/Controllers/AdminController.cs
TomHanmer29: I dont think so! I feel like theres quite a few unused imports across lots of the files, maybe it's worth raising a ticket to try and get rid of them?

web/wwwroot/js/site.js
```diff
@@ -17,3 +17,12 @@ function checkIfSelectedValueIsTarget(target){
     }
 }
 
+function copyLinkToClipboard(linkToCopy){
+    // Copy the text passed into the function plus the rest of the url
+    let fullLink = window.location.origin+"/invite/"+linkToCopy
+    navigator.clipboard.writeText(fullLink);
```
oskarwilliams378, line 23: This is actually an asynchronous function and will take a period of time to run, it may be worth putting the 2 copy button changes in a `.then()`, as a result.

web/wwwroot/js/site.js
```diff
@@ -17,3 +17,12 @@ function checkIfSelectedValueIsTarget(target){
     }
 }
 
+function copyLinkToClipboard(linkToCopy){
+    // Copy the text passed into the function plus the rest of the url
+    let fullLink = window.location.origin+"/invite/"+linkToCopy
```
oskarwilliams378, line 22: This could be a const :)

web/wwwroot/js/site.js
```diff
@@ -17,3 +17,12 @@ function checkIfSelectedValueIsTarget(target){
...
 }
 
+function copyLinkToClipboard(linkToCopy){
+    // Copy the text passed into the function plus the rest of the url
+    let fullLink = window.location.origin+"/invite/"+linkToCopy
+    navigator.clipboard.writeText(fullLink);
+    
+    // Change the button styling to show the link has been copied
+    document.getElementById("copyBtn").innerHTML = "<i class='bi bi-check'></i> invite link copied"
+    document.getElementById("copyBtn").className = "btn btn-success my-3"
```
oskarwilliams378, line 27: We could extract the copy button into its own variable here then we wouldn't have to get it twice :)

web/wwwroot/js/site.js
```diff
@@ -17,3 +17,12 @@ function checkIfSelectedValueIsTarget(target){
...
+}
```
oskarwilliams378, line 28: All the comments above would combine to something like 

```suggestion
function copyLinkToClipboard(linkToCopy) {
    // Copy the text passed into the function plus the rest of the url
    const fullLink = window.location.origin + "/invite/" + linkToCopy
    navigator.clipboard.writeText(fullLink)
        .then(() => {
                // Change the button styling to show the link has been copied
                const copyButton = document.getElementById("copyBtn")
                copyButton.innerHTML = "<i class='bi bi-check'></i> invite link copied"
                copyButton.className = "btn btn-success my-3"
            }
        );
}

```

oskarwilliams378 (changes requested):
Just a few minor comments here :)

web/Controllers/AdminController.cs
BenRamchandani: Rider has an organise imports action, feel free to just do it and raise a PR

web/wwwroot/js/site.js
TomHanmer29: Cool thanks :) I'll reopen the branch and make these changes!

PR #124, opened 2022-08-11
BNH-135&121 feat: standardise links in tables

cec200wire:


web/Views/Admin/PropertyList.cshtml
```diff
@@ -53,13 +53,20 @@
             }
             
             <td>
-                <a type="button" class="btn btn-outline-primary" value="@property.PropertyId" href="@Url.Action("ViewProperty", "Property", new { id = property.PropertyId })">
-                    <i class="bi bi-house"></i>
+                <a type="button" class="btn btn-outline-primary" value="@property.LandlordId" href="@Url.Action("Profile", "Landlord", new { id = property.LandlordId })">
```
oskarwilliams378, line 56: Do we need this value? And do we need it in the other places we are also setting this value?

oskarwilliams378 (changes requested):
Just a minor comment :)

web/Views/Admin/PropertyList.cshtml
cec200wire: We don't need that value - the routing is conveying all the information now, so setting the value with value= is unnecessary. I've removed it (here, and in a couple of other places).

oskarwilliams378: Thanks :)

oskarwilliams378 (approved):

PR #125, opened 2022-08-11
BNH-123 Fix/neatening

TomHanmer29:
Applied some changes from Oskar, used this() to await an async javascript function, small amount of refactoring such as changing lets to consts and multilining some long code


MaxTheComputerer (approved):
Looks nice!

PR #126, opened 2022-08-11
BNH-63 - unassigned landlord list

cec200wire:
Added filters to the Landlord List so that admins can filter the list by assigned/unassigned, as well as by approved/unapproved. In the process of doing so, I discovered a mostly redundant endpoint for the TenantList, so I've changed this slightly.

Also added a button to the Landlord List for admins to Create Unassigned Landlords.

### DIFF
--- a/web/Migrations/BricksAndHeartsDbContextModelSnapshot.cs
+++ b/web/Migrations/BricksAndHeartsDbContextModelSnapshot.cs
@@ -67,6 +67,9 @@ namespace BricksAndHearts.Migrations
                     b.Property<string>("InviteLink")
                         .HasColumnType("nvarchar(max)");
 
+                    b.Property<bool>("IsAssigned")
+                        .HasColumnType("bit");
+
                     b.Property<bool>("IsLandlordForProfit")
                         .HasColumnType("bit");
 
--- a/web/Migrations/BricksAndHeartsDbContextModelSnapshot.cs
+++ b/web/Migrations/BricksAndHeartsDbContextModelSnapshot.cs
@@ -67,9 +67,6 @@ namespace BricksAndHearts.Migrations
                     b.Property<string>("InviteLink")
                         .HasColumnType("nvarchar(max)");
 
-                    b.Property<bool>("IsAssigned")
-                        .HasColumnType("bit");
-
                     b.Property<bool>("IsLandlordForProfit")
                         .HasColumnType("bit");
 
--- a/web/Services/AdminService.cs
+++ b/web/Services/AdminService.cs
@@ -18,7 +18,7 @@ public interface IAdminService
     //Information Lists
     public Task<(List<UserDbModel> CurrentAdmins, List<UserDbModel> PendingAdmins)> GetAdminLists();
     public Task<List<LandlordDbModel>> GetLandlordList(LandlordListModel landlordListModel);
-    public Task<List<TenantDbModel>> GetTenantList(string[] filterArray);
+    public Task<List<TenantDbModel>> GetTenantList(TenantListFilter filter);
     
     //Invite Links
     public UserDbModel? FindUserByLandlordId(int landlordId);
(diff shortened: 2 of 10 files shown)


web/Migrations/BricksAndHeartsDbContextModelSnapshot.cs
```diff
@@ -67,6 +67,9 @@ protected override void BuildModel(ModelBuilder modelBuilder)
                     b.Property<string>("InviteLink")
                         .HasColumnType("nvarchar(max)");
 
+                    b.Property<bool>("IsAssigned")
+                        .HasColumnType("bit");
```
BenRamchandani, line 71: Is this supposed to be here?

web/Services/AdminService.cs
```diff
@@ -98,22 +97,33 @@ private async Task<List<UserDbModel>> GetPendingAdmins()
         return pendingAdmins;
     }
 
-    public async Task<List<LandlordDbModel>> GetLandlordDisplayList(string approvalStatus)
+    public async Task<List<LandlordDbModel>> GetLandlordList(string[] filterArray)
     {
-        return approvalStatus switch
+        var landlordQuery = from landlords in _dbContext.Landlords select landlords;
+        for (var currentFilterNo = 0; currentFilterNo < filterArray.Length; currentFilterNo++)
```
BenRamchandani, line 103: I know this is similar to below, but I think it would be simpler to have a view model object with bool? properties for approval status and assigned, rather than having an array where you just have to know what filter is what

Also I _think_ `from landlords in _dbContext.Landlords select landlords` is the same as just `_dbContext.Landlords`

web/ViewModels/LandlordListModel.cs
```diff
@@ -3,5 +3,6 @@ namespace BricksAndHearts.ViewModels;
 
 public class LandlordListModel
 {
-    public List<LandlordDbModel>? LandlordDisplayList { get; set; }
+    public List<LandlordDbModel>? LandlordList { get; set; }
```
BenRamchandani, line 6: Does this need to be nullable?

BenRamchandani (changes requested):

web/Migrations/BricksAndHeartsDbContextModelSnapshot.cs
cec200wire: Not really - I had minor problems getting the filtering to work (I was trying to use the function FindUserByLandlordId inside the landlordQuery.Where()), and at one stage added another field to the database. I then realised how much of a pain that would be and undid the database migrations, but those lines must have stayed around.

web/Services/AdminService.cs
cec200wire: Given that we're unlikely to be filtering the landlords by many (if any) other properties, that probably is better - it will be easier to keep track of, and won't cause a load of tedious repetition.

web/ViewModels/LandlordListModel.cs
cec200wire: I believe so, because there could (theoretically) be a scenario in which there were no landlords.

web/Services/AdminService.cs
```diff
@@ -98,22 +98,25 @@ private async Task<List<UserDbModel>> GetPendingAdmins()
...
-            _ => await _dbContext.Landlords.ToListAsync(),
+            true => _dbContext.Landlords.Where(l => l.CharterApproved == true),
+            false => _dbContext.Landlords.Where(l => l.CharterApproved == false),
+            _ => _dbContext.Landlords
         };
+        
+        landlordQuery = landlordListModel.Assigned switch
+        {
+            true => landlordQuery.Where(l => _dbContext.Users.SingleOrDefault(u => u.LandlordId == l.Id) != null),
+            false => landlordQuery.Where(l => _dbContext.Users.SingleOrDefault(u => u.LandlordId == l.Id) == null),
```
BenRamchandani, line 113: Would you mind changing these to `_dbContext.Users..Any` and `!_dbContext.Users..Any` so they don't break if we add multiple users for each landlord?

web/Views/Admin/LandlordList.cshtml
```diff
@@ -6,18 +6,32 @@
...
-    }
-</div>
+<a class="btn btn-primary" asp-controller="Landlord" asp-action="RegisterGet" asp-route-createUnassigned="true">Create Unassigned Landlord</a> <br/>
+
+<p></p>
+@using (Html.BeginForm("LandlordList", "Admin", FormMethod.Get))
+{
+    <div class="row g-3">
+        <div class="col-auto d-grid">
+            <select class="form-select" id="Approved" name="Approved">
```
BenRamchandani, line 16: Can you use `asp-for` or whatever it's called instead of specifying `name`? That way if we rename the properties on LandlordListModel these should be updated as well by Rider.

BenRamchandani (commented):
Couple of tiny things, feel free to get someone else to review and merge if I'm being slow.

web/Views/Admin/LandlordList.cshtml
cec200wire: I talked to Oskar about these forms, and he suggested rewriting them as Html.DropDownListFor() - the property name is now defined by model => model.IsApproved, so I think that should work as you suggest anyway.

web/Services/AdminService.cs
cec200wire: This section of code got slightly rewritten, and when I try changing SingleOrDefault to Any it causes issues with non-equivalent variable types - there might be a way around this, but I'm not certain it's urgent. Maybe we could figure this out in the future, if we get around to adding multiple users per landlord? It's just that I know Tom is waiting for this branch to merge in order to rebase some styling work onto it, so I don't want to delay that for too long.

web/Services/AdminService.cs
```diff
@@ -98,46 +98,50 @@ private async Task<List<UserDbModel>> GetPendingAdmins()
...
-    }
-    
-    public async Task<List<TenantDbModel>> GetTenantList()
-    {
-        return await _dbContext.Tenants.ToListAsync();
+            landlordQuery = landlordQuery.Where(l => l.CharterApproved == landlordListModel.IsApproved);
+        }
+        if (landlordListModel.IsAssigned != null)
+        {
+            bool? target = landlordListModel.IsAssigned;
```
oskarwilliams378, line 110: Do you need this variable?

web/Services/AdminService.cs
```diff
@@ -98,46 +98,50 @@ private async Task<List<UserDbModel>> GetPendingAdmins()
...
+            landlordQuery = landlordQuery.Where(l => _dbContext.Users.SingleOrDefault(u => u.LandlordId == l.Id) != null == target);
```
oskarwilliams378, line 111:
```suggestion
            landlordQuery = landlordQuery.Where(l => _dbContext.Users.Any(u => u.LandlordId == l.Id) == landlordListModel.IsAssigned);

```

oskarwilliams378 (changes requested):

web/Services/AdminService.cs
cec200wire: When I tried to write it as a single line, there seemed to be an issue with it being unable to keep track of the various == - I wasn't certain why it was happening, an thought using a separate variable might just help to simplify it slightly.

web/Services/AdminService.cs
cec200wire: Ok that does work - I think I was misunderstanding how Any worked, so was trying to use it in the wrong way.

web/Services/AdminService.cs
cec200wire: It's now been fixed - see my comment below. I think I misunderstood how Any worked (a quick google didn't help), so was using it the wrong way and thus causing the error.

web/Views/Admin/LandlordList.cshtml
BenRamchandani: Yeah that sounds fine. For the record the reason we haven't been using the helpers for selects everywhere is that as far as I can tell you can't do placeholder (i.e. a default value that you can't see in the dropdown) values with `selected disabled`, whereas you can with the tag helper things.

But that's not relevant here since `all` is an option that we want to be selectable, not just a placeholder.

oskarwilliams378:
This looks great!

PR #128, opened 2022-08-11
BNH-128 tenant filtering styling/refactor

TomHanmer29:
Put the filters into a collapsible menu and added margins


TomHanmer29:
image
image

BenRamchandani (dismissed):
This might conflict a bit with Claire's #126 FYI.

Should the arrow on the filter button flip around? Also wonder if the other button should be called "Apply" or something so it's different

TomHanmer29:
Yeah these are good points. I've looked at Claire's branch just now and I think the way she's named her refactored methods makes a bit more sense so I will wait for her to merge in first and then reapply my changes as necessary

TomHanmer29:
Yeah these are good points. I've looked at Claire's branch just now and I think the way she's named her refactored methods makes a bit more sense so I will wait for her to merge in first and then reapply my changes as necessary

TomHanmer29:
I also need to add the option to filter by e.g. "no couples", this should give me the chance to do another refactor to make everything a bit neater :)

TomHanmer29:
I have used the same object here for the filters in the property form and filters for the tenant. Not sure if this is necessarily a great idea cause it could make variable names confusing but I think it will make coding for the matching algorithm quite a bit easier.

BricksAndHearts.UnitTests/ControllerTests/Admin/AdminControllerTests.cs
```diff
@@ -1,3 +1,4 @@
+using System;
```
BenRamchandani, line 1: Presumably unused!

BenRamchandani (approved):

SleepyUser (approved):

PR #129, opened 2022-08-11
BNH-114 fix: Clear membership ID field

MaxTheComputerer:
When editing a landlord, if the charter status is changed from "Already approved" to "Not yet approved" we clear the membership ID input (which is hidden at this point).
Additionally, when a landlord is edited and the membership ID is set to null (or is already null), we un-approve the landlord's charter.

### DIFF
--- a/web/Services/LandlordService.cs
+++ b/web/Services/LandlordService.cs
@@ -46,6 +46,7 @@ public interface ILandlordService
     public bool CheckForDuplicateEmail(LandlordProfileModel editModel);
     public bool CheckForDuplicateMembershipId(LandlordProfileModel editModel);
     public Task<ApproveLandlordResult> ApproveLandlord(int landlordId, BricksAndHeartsUser user, string membershipId);
+    public Task UnapproveLandlord(int landlordId);
     public LandlordDbModel? FindLandlordWithInviteLink(string inviteLink);
 
     public Task<LinkUserWithLandlordResult> LinkExistingLandlordWithUser(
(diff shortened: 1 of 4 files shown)


web/Services/LandlordService.cs
```diff
@@ -218,6 +219,21 @@ await using (var transaction = await _dbContext.Database.BeginTransactionAsync(I
         return ILandlordService.ApproveLandlordResult.Success;
     }
 
+    public async Task UnapproveLandlord(int landlordId)
+    {
+        var landlord = await GetLandlordIfExistsFromId(landlordId);
+        if (landlord is null)
+        {
+            return;
+        }
```
BenRamchandani, line 228: Silently doing nothing is usually a bad way to handle errors because you end up with hard to diagnose bugs. So just do `Landlords.Single` and let it throw an exception if it doesn't exist.

web/Views/Landlord/_RegisterPagePartial.cshtml
```diff
@@ -103,7 +103,7 @@
             <div class="form-group my-3">
                 <label for="Select" class="form-label">Charter status</label>
                 <select class="form-select" required
-                        id="Select" onclick="checkIfSelectedValueIsTarget('true')">
+                        id="Select" onclick="checkIfSelectedValueIsTarget('true'); clearMembershipIdIfFalse()">
```
BenRamchandani, line 106: Not your change, but should the `onclick` be `onchange` really?

BenRamchandani (changes requested):

web/Services/LandlordService.cs
MaxTheComputerer: Oh yeah that's definitely a good idea

BenRamchandani (approved):

PR #132, opened 2022-08-12
BNH-109 - Homepage Restyling

cec200wire:
Added redirection to Index so that landlords/admins are displayed the relevant dashboard (admin takes priority, in the case where a user is both). Restyled the Home page for logged-out users and users who are neither landlords nor admins.
Figma design:
image

View for logged-in users:
image

View for logged-out users:
image

### DIFF
--- a/web/Views/Home/Index.cshtml
+++ b/web/Views/Home/Index.cshtml
@@ -3,30 +3,78 @@
     ViewData["Title"] = "Home Page";
 }
 
-<div class="text-center">
-    <h1 class="display-4">Bricks & Hearts</h1>
-    <p>Welcome to Bricks & Hearts.</p>
-    <p>Under construction for <a href="https://www.changeahead.org.uk/">Change Ahead</a></p>
-    <p style="max-width: 800px; margin-left: auto; margin-right: auto">This is a (work in progress) portal for landlords to manage properties they would like to make available to disadvantaged tenants looking for social housing.</p>
-    @{
-        if (Model.IsLoggedIn)
-        {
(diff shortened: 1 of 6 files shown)


SleepyUser (approved):

web/Views/Home/Index.cshtml
```diff
@@ -3,30 +3,79 @@
...
-    }
-</div>
\ No newline at end of file
+
+@if (Model.IsLoggedIn)
+{
+    <div class="container-fluid text-center">
+        <div class="row mb-4">
+            <h2>Welcome to</h2>
+            <statistics class="display-4">Bricks & Hearts</statistics>
```
BenRamchandani, line 12: What's the `<statistics>` tag about, never seen that before?

I'd expect something like

```suggestion
            <h2>Welcome to <span class="display-4">Bricks & Hearts</span></h2>
```
So a screen reader can figure out it's all part of the header

BenRamchandani (commented):
Looks good on desktop, but not so much on mobile. Can you have a go at updating it? Percentage widths often don't work as well on mobile, might have to change those. Doesn't need to look perfect, but should be usable. Chrome DevTools has a device toolbar thing to change the size.

cec200wire:
This is roughly what the mobile view looks like now - the buttons remain the correct size (rather than expanding to the full width of the page), and there is a space between them.
image

oskarwilliams378 (approved):
Other than one tiny coment approved

SleepyUser (approved):

PR #133, opened 2022-08-12
BNH-125 - Remove admins

cec200wire:
Admins can remove other admins through the admin lists; admins cannot revoke their own admin status.

### DIFF
--- a/web/Services/AdminService.cs
+++ b/web/Services/AdminService.cs
@@ -14,6 +14,7 @@ public interface IAdminService
     public void CancelAdminAccessRequest(BricksAndHeartsUser user);
     public void ApproveAdminAccessRequest(int userId);
     public void RejectAdminAccessRequest(int userId);
+    public void RemoveAdmin(int userId);
 
     //Information Lists
     public Task<(List<UserDbModel> CurrentAdmins, List<UserDbModel> PendingAdmins)> GetAdminLists();
@@ -87,6 +88,18 @@ public class AdminService : IAdminService
         }
 
         userToAdmin.HasRequestedAdmin = false;
+        _dbContext.SaveChanges();
+    }
+    
+    public void RemoveAdmin(int userId)
+    {
+        var userToUnAdmin = _dbContext.Users.SingleOrDefault(u => u.Id == userId);
+        if (userToUnAdmin == null)
+        {
+            throw new Exception($"No user found with id {userId}");
+        }
+
+        userToUnAdmin.IsAdmin = false;
         _dbContext.SaveChanges();
     }
 
--- a/web/Views/Admin/AdminList.cshtml
+++ b/web/Views/Admin/AdminList.cshtml
@@ -28,7 +28,7 @@
                     }
                 </td>
                 <td>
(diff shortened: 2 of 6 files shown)


web/Services/AdminService.cs
```diff
@@ -89,6 +90,18 @@ public void RejectAdminAccessRequest(int userId)
...
         _dbContext.SaveChanges();
     }
+    
+    public void RemoveAdmin(int userId)
+    {
+        var userToUnAdmin = _dbContext.Users.SingleOrDefault(u => u.Id == userId);
+        if (userToUnAdmin == null)
+        {
+            throw new Exception($"No user found with id {userId}");
+        }
```
MaxTheComputerer, line 100: You might be able to replace `SingleOrDefault` with `Single` to avoid throwing your own exception, because `Single` will throw an exception if there isn't exactly one record found.

web/Views/Admin/AdminList.cshtml
```diff
@@ -52,6 +53,12 @@
             <tr>
                 <td>@currentAdmin.GoogleEmail</td>
                 <td>@currentAdmin.GoogleUserName</td>
+                <td>
+                    @using (Html.BeginForm("RemoveAdmin", "Admin", new { userToRemoveId = currentAdmin.Id }, FormMethod.Post))
+                    {
+                        <button type="submit" class="btn btn-outline-danger">Remove</button>
```
MaxTheComputerer, line 59: I'm wondering if it might be clearer to have this button say "Revoke" or similar seeing as we're not removing the user, just removing their admin status. What do you think?

web/Services/AdminService.cs
cec200wire: I suspect the exception thrown by using Single gives a less useful message - I was mostly just working from the same logic as the Accept and Reject Admin Access Requests methods, which do it this way.

web/Views/Admin/AdminList.cshtml
cec200wire: That's true - I'll reword it.

web/Services/AdminService.cs
MaxTheComputerer: That's true, it's probably fine then! Alternatively, we could show a flash message for it, but this error is unlikely to ever happen so it's probably not worth the effort.

web/Views/Admin/AdminList.cshtml
```diff
@@ -52,6 +53,12 @@
             <tr>
                 <td>@currentAdmin.GoogleEmail</td>
                 <td>@currentAdmin.GoogleUserName</td>
+                <td>
+                    @using (Html.BeginForm("RemoveAdmin", "Admin", new { userToRemoveId = currentAdmin.Id }, FormMethod.Post))
+                    {
+                        <button type="submit" class="btn btn-outline-danger">Revoke Admin Rights</button>
```
MaxTheComputerer, line 59:
```suggestion
                        <button type="submit" class="btn btn-outline-danger">Revoke admin rights</button>

```
I think the designs use sentence case for all buttons

MaxTheComputerer (approved):

PR #136, opened 2022-08-15
BNH-129 - refactor flash messages

cec200wire:
Refactored the flash messages so that there is a single method, dealing with single and multiple flash messages.

### DIFF
--- a/BricksAndHearts.UnitTests/ControllerTests/Landlord/LandlordControllerTests.cs
+++ b/BricksAndHearts.UnitTests/ControllerTests/Landlord/LandlordControllerTests.cs
@@ -65,11 +65,10 @@ public class LandlordControllerTests : LandlordControllerTestsBase
 
         // Act
         await UnderTest.ApproveCharter(landlordProfile);
-        var flashMessages = UnderTest.TempData["FlashMessages"] as List<string>;
 
         // Assert
         A.CallTo(() => LandlordService.ApproveLandlord(landlordUser.Id, adminUser, "abc")).MustHaveHappened();
-        flashMessages.Should().Contain("Successfully approved landlord charter.");
+        FlashMessages.Should().Contain("Successfully approved landlord charter.");
     }
 
     [Fact]
@@ -83,11 +82,10 @@ public class LandlordControllerTests : LandlordControllerTestsBase
 
         // Act
         await UnderTest.ApproveCharter(landlordProfile);
-        var flashMessages = UnderTest.TempData["FlashMessages"] as List<string>;
 
         // Assert
         A.CallTo(() => LandlordService.ApproveLandlord(landlordUser.Id, adminUser, null!)).MustNotHaveHappened();
-        flashMessages.Should().Contain("Membership ID is required.");
+        FlashMessages.Should().Contain("Membership ID is required.");
     }
 
     [Fact]
@@ -284,7 +282,7 @@ public class LandlordControllerTests : LandlordControllerTestsBase
         result.Should().BeOfType<RedirectToActionResult>().Which.ActionName.Should().Be("Invite");
         result.Should().BeOfType<RedirectToActionResult>().Which.RouteValues.Should()
             .Contain("inviteLink", inviteLink);
-        UnderTest.TempData["FlashMessage"].Should().BeNull();
+        UnderTest.TempData["FlashMessages"].Should().BeNull();
     }
 
     [Fact]
@@ -300,11 +298,10 @@ public class LandlordControllerTests : LandlordControllerTestsBase
 
         // Act
         var result = await UnderTest.TieUserWithLandlord(inviteLink);
-        var flashMessages = UnderTest.TempData["FlashMessages"] as List<string>;
 
         // Assert   
         result.Should().BeOfType<RedirectToActionResult>().Which.ActionName.Should().Be("MyProfile");
-        flashMessages.Should()
+        FlashMessages.Should()
             .Contain($"User already registered with landlord (landlordId = {landlordUser.LandlordId})");
     }
 
@@ -321,11 +318,10 @@ public class LandlordControllerTests : LandlordControllerTestsBase
 
         // Act
         var result = await UnderTest.TieUserWithLandlord(inviteLink);
-        var flashMessages = UnderTest.TempData["FlashMessages"] as List<string>;
 
         // Assert
         result.Should().BeOfType<RedirectToActionResult>().Which.ActionName.Should().Be("MyProfile");
-        flashMessages.Should()
+        FlashMessages.Should()
             .Contain($"User {nonLandlordUser.Id} successfully linked with landlord (landlordId = {nonLandlordUser.LandlordId})");
     }
 }
--- a/BricksAndHearts.UnitTests/ControllerTests/Landlord/LandlordControllerTests.cs
+++ b/BricksAndHearts.UnitTests/ControllerTests/Landlord/LandlordControllerTests.cs
@@ -1,3 +1,4 @@
+using System.Collections.Generic;
 using System.Linq;
 using System.Threading.Tasks;
 using BricksAndHearts.Controllers;
@@ -64,10 +65,11 @@ public class LandlordControllerTests : LandlordControllerTestsBase
 
         // Act
         await UnderTest.ApproveCharter(landlordProfile);
+        var flashMessages = UnderTest.TempData["FlashMessages"] as List<string>;
 
         // Assert
         A.CallTo(() => LandlordService.ApproveLandlord(landlordUser.Id, adminUser, "abc")).MustHaveHappened();
-        UnderTest.TempData["FlashMessage"].Should().Be("Successfully approved landlord charter.");
+        flashMessages.Should().Contain("Successfully approved landlord charter.");
     }
 
     [Fact]
@@ -81,10 +83,11 @@ public class LandlordControllerTests : LandlordControllerTestsBase
 
         // Act
         await UnderTest.ApproveCharter(landlordProfile);
+        var flashMessages = UnderTest.TempData["FlashMessages"] as List<string>;
 
         // Assert
         A.CallTo(() => LandlordService.ApproveLandlord(landlordUser.Id, adminUser, null!)).MustNotHaveHappened();
-        UnderTest.TempData["FlashMessage"].Should().Be("Membership ID is required.");
+        flashMessages.Should().Contain("Membership ID is required.");
     }
 
     [Fact]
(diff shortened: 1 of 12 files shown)


web/Controllers/AbstractController.cs
```diff
@@ -16,24 +17,17 @@ protected BricksAndHeartsUser GetCurrentUser()
...
-        logger.LogInformation(flash.logInfo);
-        TempData["FlashType"] = flash.flashtype;
-        TempData["FlashMessage"] = flash.flashmessage;
-    }
-    
-    protected void FlashMultipleMessages(List<string> flashTypes, List<string> flashMessages)
-    {
-        TempData["MultipleFlashTypes"] = flashTypes;
-        TempData["MultipleFlashMessages"] = flashMessages;
+        if (TempData["FlashMessages"] == null||TempData["FlashTypes"] == null)
```
MaxTheComputerer, line 22:
```suggestion
        if (TempData["FlashMessages"] == null || TempData["FlashTypes"] == null)
```

web/Views/Shared/_Layout.cshtml
```diff
@@ -84,12 +84,18 @@
...
+                var flashMessages = TempData["FlashMessages"] as string[];
+                var flashTypes = TempData["FlashTypes"] as string[];
+                for (int i = 0; i < flashMessages!.Length; i++)
+                {
+                    <div class="alert alert-@flashTypes![i]" role="alert">
+                    @flashMessages[i]
+                    </div>
+                }
+            }
         }
```
MaxTheComputerer, line 99:
```suggestion
        @if (TempData["FlashMessages"] != null)
        {
            var flashMessages = TempData["FlashMessages"] as string[];
            var flashTypes = TempData["FlashTypes"] as string[];
            for (int i = 0; i < flashMessages!.Length; i++)
            {
                <div class="alert alert-@flashTypes![i]" role="alert">
                    @flashMessages[i]
                </div>
            }
        }
```

web/Controllers/AbstractController.cs
```diff
@@ -16,24 +17,17 @@ protected BricksAndHeartsUser GetCurrentUser()
...
+        {
+            TempData["FlashTypes"] = new List<string>{flashType};
+            TempData["FlashMessages"] = new List<string> {flashMessage};
+        }
+        else
+        {
+            (TempData["FlashTypes"] as List<string>)!.Add(flashType);
+            (TempData["FlashMessages"] as List<string>)!.Add(flashMessage);
+        }
```
oskarwilliams378, line 31: We can actually simplify this quite a lot like this, thanks to the wonders of null assigning
```suggestion
        TempData["FlashTypes"] ??= new List<string>();
        TempData["FlashMessages"] ??= new List<string>();
        
        (TempData["FlashTypes"] as List<string>)!.Add(flashType);
        (TempData["FlashMessages"] as List<string>)!.Add(flashMessage);
```

web/Views/Shared/_Layout.cshtml
```diff
@@ -84,12 +84,18 @@
...
-        {
-        @* uses warning as the default alert colour if there's no value in TempData["FlashType"] *@
-            <div class="alert alert-@(TempData["FlashType"] ?? "warning")" role="alert">
-                @TempData["FlashMessage"]
-            </div>
+        @{
+            if (TempData["FlashMessages"] != null)
+            {
+                var flashMessages = TempData["FlashMessages"] as string[];
+                var flashTypes = TempData["FlashTypes"] as string[];
```
oskarwilliams378, line 91: We can combine the casting and the checking together to be something like this :) 
```suggestion
        @if (TempData["FlashMessages"] is string[] flashMessages
            && TempData["FlashTypes"] is string[] flashTypes)
        {
```

web/Views/Shared/_Layout.cshtml
```diff
@@ -84,12 +84,18 @@
...
+                for (int i = 0; i < flashMessages!.Length; i++)
```
oskarwilliams378, line 92: could we use var not int here pls :)

web/Views/Shared/_Layout.cshtml
```diff
@@ -84,12 +84,18 @@
...
+                {
+                    <div class="alert alert-@flashTypes![i]" role="alert">
+                    @flashMessages[i]
```
oskarwilliams378, line 95: Indentation is off here :)

MaxTheComputerer (commented):
This looks much nicer now!

web/Controllers/AbstractController.cs
```diff
@@ -2,6 +2,7 @@
 
 using BricksAndHearts.Auth;
 using Microsoft.AspNetCore.Mvc;
+using Microsoft.Extensions.Azure;
```
oskarwilliams378, line 5: Don't think this is used :)

web/Controllers/AdminController.cs
```diff
@@ -164,9 +166,10 @@ public ActionResult GetInviteLink(int landlordId)
                 flashMessageBody = "Landlord already has an invite link";
             }
 
-            FlashMessage(_logger, (flashMessageBody, "success", flashMessageBody + $": {(HttpContext.Request.IsHttps ? "https" : "http")}" +
+            _logger.LogInformation(flashMessageBody);
+            AddFlashMessage("success", flashMessageBody + $": {(HttpContext.Request.IsHttps ? "https" : "http")}" +
```
oskarwilliams378, line 170: could we extract the flashMessage body out of this call to add message and then pass it in, something like 
```suggestion
            _logger.LogInformation(logMessage);
            var flashMessage = logMessage +
                $": {(HttpContext.Request.IsHttps ? "https" : "http")}://{HttpContext.Request.Host}/invite/{inviteLink}";
            AddFlashMessage("success", flashMessage);
```

web/Services/CsvImportService.cs
```diff
@@ -63,7 +63,7 @@ public CsvImportService(BricksAndHeartsDbContext dbContext, ILogger<CsvImportSer
         return (flashTypes, flashMessages);
     }
 
-    public async Task<(List<string> FlashTypes, List<string> FlashMessages)> ImportTenants(IFormFile csvFile, (List<string> flashTypes, List<string> flashMessages) flashResponse)
+    public async Task<(List<string> FlashTypes, List<string> FlashMessages)> ImportTenants(IFormFile csvFile)
     {
         CsvFileDescription csvFileDescription = new CsvFileDescription
```
oskarwilliams378, line 68: I know this wasn't you, but can we turn all these into `var`s please :)

BricksAndHearts.UnitTests/ControllerTests/Landlord/LandlordControllerTests.cs
```diff
@@ -297,11 +300,12 @@ public async void TieUserWithLandlord_WithLandlord_RedirectToProfile()
 
         // Act
         var result = await UnderTest.TieUserWithLandlord(inviteLink);
+        var flashMessages = UnderTest.TempData["FlashMessages"] as List<string>;
```
oskarwilliams378, line 303: On line 287 (above this but I cannot comment there) we still have the case where we are looking for `UnderTest.TempData["FlashMessage"]` (not plural), could you please change that too :)

BricksAndHearts.UnitTests/ControllerTests/Landlord/LandlordControllerTests.cs
```diff
@@ -81,10 +83,11 @@ public async void ApproveCharter_WithoutMembershipId_DisplaysErrorMessage()
 
         // Act
         await UnderTest.ApproveCharter(landlordProfile);
+        var flashMessages = UnderTest.TempData["FlashMessages"] as List<string>;
```
oskarwilliams378, line 86: As we are calling the same thing everytime to access the flashMessages, we could add something to the TestsBase, something like 
```
protected IEnumerable<string>? FlashMessages => UnderTest.TempData["FlashMessages"] as List<string>;
```

This would give us access to the FlashMessages without having to add them into all the Acts

oskarwilliams378 (changes requested):
Looks much much nicer now! Just a couple things to add :)

web/Controllers/AbstractController.cs
cec200wire: I distinctly remember deleting that, it must have crept back in a rebase.

web/Services/CsvImportService.cs
cec200wire: I assume you mean the CsvFileDescription, CsvContext etc. at the top of this method, since trying to not specify the output types of the method makes the IDE complain.

BricksAndHearts.UnitTests/ControllerTests/Landlord/LandlordControllerTests.cs
cec200wire: Yes, I hadn't noticed that one because the unit test wasn't failing!

web/Services/CsvImportService.cs
oskarwilliams378: Yep, things like
```
var csvFileDescription = new CsvFileDescription
```

BricksAndHearts.UnitTests/ControllerTests/Landlord/LandlordControllerTests.cs
```diff
@@ -1,3 +1,4 @@
+using System.Collections.Generic;
```
oskarwilliams378, line 1: Not needed :)

BricksAndHearts.UnitTests/ControllerTests/Landlord/LandlordControllerTests.cs
```diff
@@ -281,7 +282,7 @@ public async void TieUserWithLandlord_WithNonExistentLink_RedirectToInvite()
         result.Should().BeOfType<RedirectToActionResult>().Which.ActionName.Should().Be("Invite");
         result.Should().BeOfType<RedirectToActionResult>().Which.RouteValues.Should()
             .Contain("inviteLink", inviteLink);
-        UnderTest.TempData["FlashMessage"].Should().BeNull();
+        UnderTest.TempData["FlashMessages"].Should().BeNull();
```
oskarwilliams378, line 285: Should this just use the FlashMessages?

oskarwilliams378 (approved):

BricksAndHearts.UnitTests/ControllerTests/Landlord/LandlordControllerTests.cs
cec200wire: That would be another one I missed because the test was passing (I'm now searching to check that all singular instances of "FlashMessage" are dealt with!)

cec200wire: Having just checked, that seems to have already been changed so all is fine.

PR #137, opened 2022-08-15
BNH-149 Style property details page

MaxTheComputerer:
The individual "view property" details page has been styled to match the [designs](https://www.figma.com/file/Sw8in06M3r8s78L5gafePu/CHANGEAHEAD?node-id=74%3A337).

- Clicking on any of the images shows a modal with a carousel containing all images for that property
- Images can be uploaded and deleted from the modal
- Images have a placeholder in case they are slow to load
- The edit icons currently just go to the start of the edit property form, this should probably be changed when [BNH-132](https://softwiretech.atlassian.net/browse/BNH-132) is completed
- View matches button also needs linking to the matching pages when done

image
image
image
image
image

(Branch has incorrect number)


web/ViewModels/AddressModel.cs
```diff
@@ -23,4 +24,25 @@ public class AddressModel
...
+        {
+            address.Append($", {AddressLine3}");
+        }
+
+        if (TownOrCity != null)
+        {
+            address.Append($", {TownOrCity}");
+        }
+
+        return address.ToString();
```
oskarwilliams378, line 46: This could be simplified down a bit, we could use ternaries and do something like
```suggestion
        var address = AddressLine1;
        address += AddressLine2 != null ? $", {AddressLine2}" : "";
        address += AddressLine3 != null ? $", {AddressLine3}" : "";
        address += TownOrCity != null ? $", {TownOrCity}" : "";

        return address;
```

web/ViewModels/PropertyDetailsViewModel.cs
```diff
@@ -0,0 +1,7 @@
+﻿namespace BricksAndHearts.ViewModels;
+
+public class PropertyDetailsViewModel
+{
+    public PropertyViewModel Property { get; set; } = null!;
```
oskarwilliams378, line 5: Theres a bit of a counter pattern here that I am not too happy with, what you are doing here is saying that the Property should not be null (as the type is `PropertyViewModel` not `PropertyViewModel?`), but then immediately set it to null. I think we should either have the property of this class be nullable, or have us set it to a real value to start, probably the former over the latter.

web/Views/Property/ViewProperty.cshtml
```diff
@@ -1,132 +1,338 @@
...
+</div>
+
+<div class="row property-img-row">
+    @if (Model.Images.Count > 0)
+    {
+        <div class="col-sm my-3">
+            <a href="#" class="property-img-lg placeholder-glow" data-bs-toggle="modal" data-bs-target="#imageModal">
+                <div class="placeholder w-100 h-100" id="image0Placeholder"></div>
+                <img src="@Model.Images[0].FileUrl" alt="@Model.Images[0].FileName" id="image0" style="display: none" onload="hidePlaceholder('#image0')">
+            </a>
```
oskarwilliams378, line 63: All these image links look like they could be extracted into a function with maybe 3 inputs, I think that may makeit slightly nicer :)

web/Views/Property/ViewProperty.cshtml
```diff
@@ -1,132 +1,338 @@
...
+        <div class="card-body">
+            <p class="collapse property-description font-smallBody" id="descriptionCollapse">@(Model.Property.Description ?? "Missing description")</p>
+            <a class="collapsed font-smallBody" role="button" data-bs-toggle="collapse" href="#descriptionCollapse"></a>
         </div>
     </div>
-    <div class="row mb-3">
-        <div class="col-4">
-            <h4>Property details</h4>
+</div>
+
```
oskarwilliams378, line 112: Double new line, pls remove :)

web/Views/Property/ViewProperty.cshtml
```diff
@@ -1,132 +1,338 @@
...
-        Delete Property
-    </button>
-    <br/>
-    <br/>
-    @if (User.IsInRole("Admin"))
-    {
-        @Html.ActionLink("Get Public View Link", "GetPublicViewLink", "Property", new { propertyId = Model.PropertyId }, 
-            new { @class = "btn btn-primary" })   
-    }
 </div>
```
oskarwilliams378, line 217: Double new line, pls remove :)

web/Views/Property/ViewProperty.cshtml
```diff
@@ -1,132 +1,338 @@
...
+        <a asp-action="EditProperty" asp-controller="Property" class="btn btn-large btn-outline-primary d-table-cell">Edit property</a>
+    </div>
+    <div class="col-auto">
+        <button type="button" class="btn btn-large btn-danger" data-bs-toggle="modal" data-bs-target="#deletePropertyModal">
+            <i class="bi bi-trash"></i>
+            Delete property
+        </button>
+    </div>
+</div>
+
```
oskarwilliams378, line 242: Double new line, pls remove :)

web/Views/Property/ViewProperty.cshtml
```diff
@@ -1,132 +1,338 @@
-﻿@using Microsoft.AspNetCore.Mvc.TagHelpers
-@model PropertyViewModel
+﻿@model PropertyDetailsViewModel
```
oskarwilliams378, line 1: This file is very very long, it may benefit from being broken into some partials to just separate out some of the complexity and noise :)

web/wwwroot/css/site.css
```diff
@@ -347,6 +332,12 @@ th {
     font-weight: normal;
 }
 
+.table-maxheight {
+    overflow-y: scroll;
+    overflow-x: auto;
+    max-height: 200px;
+}
+
```
oskarwilliams378, line 340: Double new line, pls remove :)

oskarwilliams378 (changes requested):

web/Controllers/PropertyController.cs
```diff
@@ -16,7 +16,8 @@ public class PropertyController : AbstractController
     private readonly ILogger<PropertyController> _logger;
     private readonly IAzureStorage _azureStorage;
 
-    public PropertyController(IPropertyService propertyService, IAzureMapsApiService azureMapsApiService, ILogger<PropertyController> logger, IAzureStorage azureStorage, IPostcodeApiService postcodeApiService)
+    public PropertyController(IPropertyService propertyService, IAzureMapsApiService azureMapsApiService,
+        ILogger<PropertyController> logger, IAzureStorage azureStorage, IPostcodeApiService postcodeApiService)
```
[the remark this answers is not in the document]
MaxTheComputerer: @cec200wire is going to use the postcode service so it might make sense to keep it here for now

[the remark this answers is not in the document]
oskarwilliams378: Sure, can we multiline the constructor call though still :)

web/Views/Property/Details/_TenantPreferences.cshtml
```diff
@@ -0,0 +1,52 @@
...
+@functions
+{
+    string GetPreferenceFont(bool? preference)
+    {
+        return preference == true ? "" : "font-orange";
+    }
+
+    string GetPreferenceIcon(bool? preference)
+    {
+        return preference ?? false ? "bi-check-lg" : "bi-x-lg";
```
oskarwilliams378, line 12: could we make this the same as above

oskarwilliams378 (approved):
One minor change, but approved as you can merge after doing it :)

web/Views/Property/Details/_TenantPreferences.cshtml
MaxTheComputerer: Oh yeah good spot!

TomHanmer29 (approved):

PR #138, opened 2022-08-16
BNH 143: Simple matching algorithm

TomHanmer29:
Added a simple matching "algorithm" according to the filters we have for tenant sorting. Once the tenant distance ticket is done, this can be sorted by distance (currently sorted by name). Added 2 buttons to get to this page, one from the admin property table and one from the individual property page but only if the user is an admin.
image
image
image
Also slightly simplified / refactored the filtering to be more in line with the landlord's requirements. Previously, all filters were exclusive (i.e. having 2 filters on showed the intersection of those 2 filters. Now filters are inclusive for the landlord type (i.e. shows the union of those 2 filters) and exclusive for everything not landlord type based on the filter page, and are inclusive everywhere for the matching algorithm.

### DIFF
--- a/web/Services/AdminService.cs
+++ b/web/Services/AdminService.cs
@@ -139,52 +139,61 @@ public class AdminService : IAdminService
         return await landlordQuery.ToListAsync();
     }
 
-    private IQueryable<TenantDbModel> GetFilteredTenantQuery(HousingRequirementModel filters)
+    private IQueryable<TenantDbModel> GetFilteredTenantQuery(HousingRequirementModel filters, bool filterOrMatching)
     {
         var tenantQuery = from tenants in _dbContext.Tenants select tenants;
-        var currentFilterNo = -1;
-        foreach (var filter in filters)
+        var filterList = filters.GetList();
+        if (filterList.SequenceEqual(new HousingRequirementModel().GetList())) return tenantQuery;
+        tenantQuery = tenantQuery.Where(t => (t.Type == "Single" && filterList[0] == true) ||
+                                             (t.Type == "Couple" && filterList[1] == true) ||
+                                             (t.Type == "Family" && filterList[2] == true) ||
+                                             (filterList[0] != true && filterList[1] != true && filterList[2] != true));
+        /*the above are INCLUSIVE filters*/
+        if (filterOrMatching)
         {
-            currentFilterNo++;
-            if (filter != null)
+            if (filterList[3]==false)
             {
-                tenantQuery = currentFilterNo switch
-                {
-                    0 => filter switch
-                    {
-                        true => tenantQuery.Where(t => t.Type == "Single"),
-                        false => tenantQuery.Where(t => t.Type != "Single")
-                    },
-                    1 => filter switch
-                    {
-                        true => tenantQuery.Where(t => t.Type == "Couple"),
-                        false => tenantQuery.Where(t => t.Type != "Couple")
-                    },
-                    2 => filter switch
-                    {
-                        true => tenantQuery.Where(t => t.Type == "Family"),
-                        false => tenantQuery.Where(t => t.Type != "Family")
-                    },
-                    3 => tenantQuery.Where(t => t.HasPet == filter),
-                    4 => tenantQuery.Where(t => t.ETT == filter),
-                    5 => tenantQuery.Where(t => t.UniversalCredit == filter),
-                    6 => tenantQuery.Where(t => t.HousingBenefits == filter),
-                    7 => tenantQuery.Where(t => t.Over35 == filter),
-                    _ => tenantQuery
-                };
+                tenantQuery = tenantQuery.Where(t => t.HasPet == false);
             }
+
+            if (filterList[4]==false)
+            {
+                tenantQuery = tenantQuery.Where(t => t.ETT == false);
+            }
+
+            if (filterList[5]==false)
+            {
+                tenantQuery = tenantQuery.Where(t => t.UniversalCredit == false);
+            }
+
+            if (filterList[6]==false)
+            {
+                tenantQuery = tenantQuery.Where(t => t.HousingBenefits == false);
+            }
+
+            if (filterList[7]==false)
+            {
+                tenantQuery = tenantQuery.Where(t => t.Over35 == false);
+            }
+            return tenantQuery;
         }
-        return tenantQuery;
+        /*Above are INCLUSIVE filters for the matching page*/
+        return tenantQuery.Where(t => (filterList[3] == null || t.HasPet == filterList[3]) &&
+                                      (filterList[4] == null || t.ETT == filterList[4]) &&
+                                      (filterList[5] == null || t.UniversalCredit == filterList[5]) &&
+                                      (filterList[6] == null || t.HousingBenefits == filterList[6]) &&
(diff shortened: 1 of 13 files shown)


web/Views/Admin/TenantMatchList.cshtml
```diff
@@ -0,0 +1,65 @@
...
+        var flashMessages = TempData["MultipleFlashMessages"] as string[];
+        var flashTypes = TempData["MultipleFlashTypes"] as string[];
+        for (int i = 0; i < flashMessages!.Length; i++)
+        {
+            <div class="alert alert-@flashTypes![i]" role="alert">
+            @flashMessages[i]
+            </div>
+        }
+    }
+}
```
cec200wire, line 20: if (TempData["MultipleFlashMessages"] != null) section here shouldn't be necessary any more.

web/Services/AdminService.cs
```diff
@@ -136,42 +139,63 @@ public async Task<List<LandlordDbModel>> GetLandlordList(LandlordListModel landl
...
+            return tenantQuery;
         }
+        /*Above are INCLUSIVE filters for the matching page*/
+        return tenantQuery.Where(t => (filterList[3] == null || t.HasPet == filterList[3]) &&
+                                      (filterList[4] == null || t.ETT == filterList[4]) &&
+                                      (filterList[5] == null || t.UniversalCredit == filterList[5]) &&
+                                      (filterList[6] == null || t.HousingBenefits == filterList[6]) &&
+                                      (filterList[7] == null || t.Over35 == filterList[7]));
+        /*Above are EXCLUSIVE filters for the filters page*/
+    }
```
oskarwilliams378, line 187:
```
        var tenantQuery = _dbContext.Tenants.AsQueryable();

        if (filter.AllFiltersAreNull) return tenantQuery;
        
        tenantQuery = tenantQuery.Where(t => (t.Type == "Single" && filter.AcceptsSingleTenant == true) 
            || (t.Type == "Couple" && filter.AcceptsCouple == true)
            || (t.Type == "Family" && filter.AcceptsFamily == true)
            || (filter.AcceptsSingleTenant != true && filter.AcceptsCouple != true && filter.AcceptsFamily != true));
        /*These are INCLUSIVE filters*/

        if (!isMatching)
        {
            /*These are EXCLUSIVE filters for the filters page*/
            return tenantQuery.Where(t => (filter.AcceptsPets == null || t.HasPet == filter.AcceptsPets)
                && (filter.AcceptsNotEET == null || t.ETT == filter.AcceptsNotEET) 
                && (filter.AcceptsCredit == null || t.UniversalCredit == filter.AcceptsCredit)
                && (filter.AcceptsBenefits == null || t.HousingBenefits == filter.AcceptsBenefits)
                && (filter.AcceptsOver35 == null || t.Over35 == filter.AcceptsOver35));
        }

        /*These are INCLUSIVE filters for the matching page*/
        if (filter.AcceptsPets == false)
        {
            tenantQuery = tenantQuery.Where(t => t.HasPet == false);
        }

        if (filter.AcceptsNotEET == false)
        {
            tenantQuery = tenantQuery.Where(t => t.ETT == false);
        }

        if (filter.AcceptsCredit == false)
        {
            tenantQuery = tenantQuery.Where(t => t.UniversalCredit == false);
        }

        if (filter.AcceptsBenefits == false)
        {
            tenantQuery = tenantQuery.Where(t => t.HousingBenefits == false);
        }

        if (filter.AcceptsOver35 == false)
        {
            tenantQuery = tenantQuery.Where(t => t.Over35 == false);
        }

        return tenantQuery;
```

web/Services/AdminService.cs
```diff
@@ -136,42 +139,63 @@ public async Task<List<LandlordDbModel>> GetLandlordList(LandlordListModel landl
         return await landlordQuery.ToListAsync();
     }
 
-    public async Task<List<TenantDbModel>> GetTenantList(HousingRequirementModel filters)
+    private IQueryable<TenantDbModel> GetFilteredTenantQuery(HousingRequirementModel filters, bool filterOrMatching)
```
oskarwilliams378, line 142: So we have all these values as bools and we are then converting them into a list. I am not exactly sure why this is happening here, instead of just using the actual boolean properties on the model with defaults as false, below is an example of how this would look.

Also the variable name `filterOrMatching` doesn't actually tell me whether it is a filter or matching, could you please rename :)

oskarwilliams378: Oh and can we please put the private methods below the public methods

web/Services/MailService.cs
```diff
@@ -34,11 +36,13 @@ public MailService(IOptions<EmailConfigOptions> config, ILogger<MailService> log
     public async Task SendMsg(
         string msgBody,
```
oskarwilliams378, line 37: Do we need to have msg prefixing all these variables?

web/Services/MailService.cs
```diff
@@ -10,13 +10,15 @@ public interface IMailService
     public Task SendMsg(
         string msgBody,
         string subject,
+        List<string>? msgToAddress,
```
oskarwilliams378, line 13: should these be something like `toAddresses`?

web/ViewModels/TenantMatchListModel.cs
```diff
@@ -0,0 +1,8 @@
+﻿using BricksAndHearts.Database;
+namespace BricksAndHearts.ViewModels;
+
+public class TenantMatchListModel
+{
+    public List<TenantDbModel>? TenantList { get; set; }
+    public PropertyViewModel CurrentProperty { get; set; } = null!;
```
oskarwilliams378, line 7: This is a bit of an anti pattern here, we are seemingly saying that the `CurrentProperty` is not null, as its not `PropertyViewModel?`, but then we are setting it to be equal to null. Either the type should change to nullable (and then we don't need to instantiate it with null), or we should be instantiating it

web/Views/Admin/TenantMatchList.cshtml
```diff
@@ -0,0 +1,50 @@
...
+            <td>@tenant.Type</td>
+            <td>@tenant.HasPet</td>
+            <td>@tenant.ETT</td>
+            <td>@tenant.UniversalCredit</td>
+            <td>@tenant.HousingBenefits</td>
+            <td>@tenant.Over35</td>
+            <td>
+                @using (Html.BeginForm("SendMatchLinkEmail", "Admin", new { propertyLink = Model.CurrentProperty.Description, /*TODO: change this from description to link*/ 
+                                                                                                        tenantEmail = tenant.Email, 
+                                                                                                        currentPropertyId = Model.CurrentProperty.PropertyId }, FormMethod.Post))
```
oskarwilliams378, line 42: indentation has gone very far here. I think we can add some more new lines in here to make it tidier :)
```suggestion@using (Html.BeginForm("SendMatchLinkEmail", "Admin", new
                {
                    propertyLink = Model.CurrentProperty.Description, /*TODO: change this from description to link*/ 
                    tenantEmail = tenant.Email,
                    currentPropertyId = Model.CurrentProperty.PropertyId
                }, FormMethod.Post))
```

oskarwilliams378 (changes requested):

oskarwilliams378 (approved):

PR #140, opened 2022-08-16
BNH-142 - Sort tenants by location

cec200wire:
Functionality added to sort tenants based on their proximity to a given postcode.
Currently, any tenant who does not have a postcode is excluded from the resulting list - this should probably be changed, to consistently place such tenants either at the beginning or end of the list. As this issue is intertwined with matching, it may make sense for the way it is resolved to depend on whether the tenants are being sorted for matching purposes or simply to be displayed, and as such it can be dealt with once BNH-143 is merged (either on this branch, or on a new one which integrates the two halves of the matching process - location and requirements)


web/Services/CsvImportService.cs
```diff
@@ -73,10 +77,12 @@ public async Task<(List<string> FlashTypes, List<string> FlashMessages)> ImportT
         };
         var csvContext = new CsvContext();
         var streamReader = new StreamReader(csvFile.OpenReadStream());
-        IEnumerable<TenantUploadModel> tenantUploadList = csvContext.Read<TenantUploadModel>(streamReader, csvFileDescription);
-        
+        IEnumerable<TenantUploadModel> tenantUploadList =
```
oskarwilliams378, line 80: A few places for var usage in this file :)

web/Services/CsvImportService.cs
```diff
@@ -88,84 +94,143 @@ await using (var transaction = await _dbContext.Database.BeginTransactionAsync(I
...
+        }
+
+        await _postcodeService.BulkAddPostcodesToDatabaseIfAbsent(postcodes);
+        return (flashTypes, flashMessages);
+    }
+    
+    private (List<string> flashTypes, List<string> flashMessages, string postcode) AddTenantToDb(TenantUploadModel tenant, int lineNo)
+    {
+        List<string> flashTypes = new(), flashMessages = new();
+        string postcode = "";
```
oskarwilliams378, line 116: var :)

web/Services/CsvImportService.cs
```diff
@@ -88,84 +94,143 @@ await using (var transaction = await _dbContext.Database.BeginTransactionAsync(I
...
+            _logger.LogWarning($"Name is missing from record {tenant.Email}.");
+            flashTypes.Add("danger");
+            flashMessages.Add($"Name is missing from record on line {lineNo} (email address provided is {tenant.Email}). This record has not been added to the database. Please add a name to this tenant in order to import their information.");
+        }
+        else
+        {
+            TenantDbModel dbTenant = new TenantDbModel();
+            dbTenant.Name = tenant.Name;
+            dbTenant.Email = tenant.Email;
+            dbTenant.Phone = tenant.Phone;
```
oskarwilliams378, line 128:
```suggestion
            var dbTenant = new TenantDbModel
            {
                Name = tenant.Name,
                Email = tenant.Email,
                Phone = tenant.Phone
            };
```

web/Services/CsvImportService.cs
```diff
@@ -88,84 +94,143 @@ await using (var transaction = await _dbContext.Database.BeginTransactionAsync(I
...
+                    flashTypes.Add("danger");
+                    flashMessages.Add(InvalidInputMessage(tenant.Name, "Postcode", tenant.Postcode!));
                 }
-                _dbContext.SaveChanges();
             }
-            await transaction.CommitAsync();
+
+            dbTenant.Type = tenant.Type;
+
+            if (Boolean.TryParse(tenant.HasPet, out bool boolHasPet) || tenant.HasPet is null)
```
oskarwilliams378, line 146: You can do similar to this below as well

```suggestion
            if (bool.TryParse(tenant.HasPet, out var boolHasPet) || tenant.HasPet is null)

```

web/Services/CsvImportService.cs
```diff
@@ -88,84 +94,143 @@ await using (var transaction = await _dbContext.Database.BeginTransactionAsync(I
...
+    public (List<string> FlashTypes, List<string> FlashMessages) AddLatLonToTenantDb()
+    {
+        List<string> flashTypes = new(), flashMessages = new();
+        var tenantsWithPostcodesToConvert =
+            _dbContext.Tenants.Where(t => t.Postcode != null && t.Lat == null && t.Lon == null).ToList();
+        
+        foreach (var tenant in tenantsWithPostcodesToConvert)
+        {
+            var postcodeDbModel = _dbContext.Postcodes.FirstOrDefault(p => p.Postcode == tenant.Postcode);
+            if (postcodeDbModel == null || postcodeDbModel.Lat == null || postcodeDbModel.Lon == null)
```
oskarwilliams378, line 221: This can be merged into two :) 
```suggestion
            if (postcodeDbModel?.Lat == null || postcodeDbModel.Lon == null)

```

web/Services/PostcodeService.cs
```diff
@@ -0,0 +1,176 @@
...
+
+namespace BricksAndHearts.Services;
+
+public interface IPostcodeService
+{
+    public string FormatPostcode(string postcode);
+    public Task AddSinglePostcodeToDatabaseIfAbsent(string postcode);
+    public Task BulkAddPostcodesToDatabaseIfAbsent(List<string> postcodes);
+}    
+
```
oskarwilliams378, line 15: extra new line

web/Services/PostcodeService.cs
```diff
@@ -0,0 +1,176 @@
...
+        }
+        else
+        {
+            postcode = "";
+        }
+
+        return postcode;
+    }
+
+    private void AddPostcodeToPostcodeDb(PostcodeioResponseModel postcodeResponse)
```
oskarwilliams378, line 45: Can we move private methods below public :)

web/Services/PostcodeService.cs
```diff
@@ -0,0 +1,176 @@
...
+                @"([Gg][Ii][Rr] 0[Aa]{2})|((([A-Za-z][0-9]{1,2})|(([A-Za-z][A-Ha-hJ-Yj-y][0-9]{1,2})|(([A-Za-z][0-9][A-Za-z])|([A-Za-z][A-Ha-hJ-Yj-y][0-9][A-Za-z]?))))\s?[0-9][A-Za-z]{2})"))
+        {
+            postcode = Regex.Replace(postcode, @"^(\S+?)\s*?(\d\w\w)$", "$1 $2").ToUpper();
+        }
+        else
+        {
+            postcode = "";
+        }
+
+        return postcode;
```
oskarwilliams378, line 42: I think this may be a little nicer if we extract out that regex as a const, then we can use it in a ternary :) 
```suggestion
        const string postcodePattern = @"([Gg][Ii][Rr] 0[Aa]{2})|((([A-Za-z][0-9]{1,2})|(([A-Za-z][A-Ha-hJ-Yj-y][0-9]{1,2})|(([A-Za-z][0-9][A-Za-z])|([A-Za-z][A-Ha-hJ-Yj-y][0-9][A-Za-z]?))))\s?[0-9][A-Za-z]{2})"
        return Regex.IsMatch(postcode, postcodePattern)
            ? Regex.Replace(postcode, @"^(\S+?)\s*?(\d\w\w)$", "$1 $2").ToUpper() 
            : "";

```

web/Services/PostcodeService.cs
```diff
@@ -0,0 +1,176 @@
...
+    }
+
+    private void AddPostcodeToPostcodeDb(PostcodeioResponseModel postcodeResponse)
+    {
+        if (postcodeResponse.Result == null || postcodeResponse.Result.Lat == null || postcodeResponse.Result.Lon == null || postcodeResponse.Result.Postcode == null)
```
oskarwilliams378, line 47: We can merge in the initial null check again :)

web/Controllers/AdminController.cs
```diff
@@ -247,17 +247,24 @@ public async Task<ActionResult> ImportTenants(IFormFile csvFile)
...
         {
             _logger.LogInformation("Successfuly imported all tenant data.");
             AddFlashMessage("success", "Successfully imported all tenant data.");
         }
+        
+        flashResponse =  _csvImportService.AddLatLonToTenantDb();
+        for (int i = 0; i < flashResponse.FlashMessages.Count(); i++)
+        {
+            AddFlashMessage(flashResponse.FlashTypes[i],  flashResponse.FlashMessages[i]);
+        }
```
MaxTheComputerer, line 266: Could this be done as part of the main `ImportTenants` method?

web/Services/PostcodeService.cs
```diff
@@ -0,0 +1,176 @@
...
+                Postcode = postcodeResponse.Result.Postcode,
+                Lat = postcodeResponse.Result.Lat,
+                Lon = postcodeResponse.Result.Lon,
+            };
+            _dbContext.Postcodes.Add(postcodeDbModel);
+        }
+        _dbContext.SaveChanges();
+    }
+    
+    public async Task AddSinglePostcodeToDatabaseIfAbsent(string postcode)
```
oskarwilliams378, line 64: Is there any difference between using this and using the BulkAdd with just a list of length one? 

If not, this method could just pass to the bulk add a list of length 1

web/Views/Admin/TenantList.cshtml
```diff
@@ -11,6 +11,13 @@
         <i class="bi bi-arrow-90deg-down"></i> Filter
     </a>
 </div>
+<div class="col-8">
+    @using (Html.BeginForm("SortTenantsByLocation", "Tenant", FormMethod.Get))
+    {
+        <input class="form-control" id="postcode" name="postcode" required placeholder="Enter postcode to sort by location" pattern="([Gg][Ii][Rr] 0[Aa]{2})|((([A-Za-z][0-9]{1,2})|(([A-Za-z][A-Ha-hJ-Yj-y][0-9]{1,2})|(([A-Za-z][0-9][A-Za-z])|([A-Za-z][A-Ha-hJ-Yj-y][0-9][A-Za-z]?))))\s?[0-9][A-Za-z]{2})"/>
+        <button type="submit" class="btn custom-btn-primary"> Sort </button>
```
MaxTheComputerer, line 18: It would be better to use padding to space out the text in this button instead of using spaces

web/Services/PostcodeService.cs
```diff
@@ -0,0 +1,176 @@
...
+            return new PostcodeioResponseModel();
+        }
+        var postcodeResponse = JsonConvert.DeserializeObject<PostcodeioResponseModel>(responseBody);
+        return postcodeResponse!;
+    }
+
+    public async Task BulkAddPostcodesToDatabaseIfAbsent(List<string> postcodes)
+    {
+        var postcodesToLookup = new List<string>();
+        foreach (string postcode in postcodes)
```
oskarwilliams378, line 108: var

web/Services/PostcodeService.cs
```diff
@@ -0,0 +1,176 @@
...
+        {
+            if (!_dbContext.Postcodes.Any(p => p.Postcode == postcode) && !postcodesToLookup.Any(p => p == postcode))
```
oskarwilliams378, line 110: we could make these Alls instead
```suggestion
            if (_dbContext.Postcodes.All(p => p.Postcode != postcode) && postcodesToLookup.All(p => p != postcode))

```

web/Services/PostcodeService.cs
```diff
@@ -0,0 +1,176 @@
...
+            {
+                postcodesToLookup.Add(postcode);
+            }
+
+            if (postcodesToLookup.Count() == 100)
```
oskarwilliams378, line 115: Count is a property as well so we can remove the `()` (here and below)

web/Services/PostcodeService.cs
```diff
@@ -0,0 +1,176 @@
...
+            {
+                await BulkGetLatLonForPostcodes(postcodesToLookup);
```
oskarwilliams378, line 117: instead of awaiting these each time, we can make a list of tasks and add these onto it, only awaiting all of them together at the end, that would prevent any blocking :)

web/Services/PostcodeService.cs
```diff
@@ -0,0 +1,176 @@
...
+            await BulkGetLatLonForPostcodes(postcodesToLookup);
+        }
+    }
+
+    private async Task BulkGetLatLonForPostcodes(List<string> postcodesToLookup)
+    {
+        var bulkResponseBody = await MakeBulkApiRequestToPostcodeApi(postcodesToLookup);
+        var postcodeResponse = TurnBulkResponseBodyToModel(bulkResponseBody);
+
+        foreach (PostcodeioResponseModel response in postcodeResponse)
```
oskarwilliams378, line 133: var

web/Services/PostcodeService.cs
```diff
@@ -0,0 +1,176 @@
...
+
+    private List<PostcodeioResponseModel> TurnBulkResponseBodyToModel(string bulkResponseBody)
+    {
+        var postcodeDbModelList = new List<PostcodeioResponseModel>();
+        if (string.IsNullOrEmpty(bulkResponseBody))
+        {
+            return postcodeDbModelList;
+        }
+        JObject jObjectResponse = JObject.Parse(bulkResponseBody);
+        IList<JToken> postcodeList = jObjectResponse["result"]!.ToList();
```
oskarwilliams378, line 168: vars

web/Services/PostcodeService.cs
```diff
@@ -0,0 +1,176 @@
...
+        {
+            _logger.LogWarning("Message :{Message} ", e.Message);
+        }
+
+        return bulkResponseBody;
+    }
+
+    private List<PostcodeioResponseModel> TurnBulkResponseBodyToModel(string bulkResponseBody)
+    {
+        var postcodeDbModelList = new List<PostcodeioResponseModel>();
```
oskarwilliams378, line 162: I dont think this variable name is accurate :)

web/Services/PropertyService.cs
```diff
@@ -1,3 +1,4 @@
+using System.Text.RegularExpressions;
```
oskarwilliams378, line 1: Not needed :)

web/Services/TenantService.cs
```diff
@@ -0,0 +1,53 @@
...
+{
+    private readonly BricksAndHeartsDbContext _dbContext;
+    private readonly IPostcodeService _postcodeService;
+
+    public TenantService(BricksAndHeartsDbContext dbContext, IPostcodeService postcodeService)
+    {
+        _dbContext = dbContext;
+        _postcodeService = postcodeService;
+    }
+
```
oskarwilliams378, line 21: extra new line

web/Services/TenantService.cs
```diff
@@ -0,0 +1,53 @@
...
+    
+    public async Task<List<TenantDbModel>?> SortTenantsByLocation(string postalCode, int page, int perPage)
+    {
+        var postcode = _postcodeService.FormatPostcode(postalCode);
+        await _postcodeService.AddSinglePostcodeToDatabaseIfAbsent(postcode);
+        var model = _dbContext.Postcodes.SingleOrDefault(p => p.Postcode == postcode);
```
oskarwilliams378, line 27: Could we call this something more descriptive please :)

oskarwilliams378 (changes requested):

web/Services/PostcodeService.cs
cec200wire: I originally put the private methods below the ones which called them (except for one which I called twice, which I put higher up) - I found it easier for keeping track of the logic while writing the code, but I suppose now it's done it would be easier to have the public methods clearly visible at the top.

web/Services/PostcodeService.cs
cec200wire: That is nicer - I'm not particularly familiar with that way of writing things, hence why I did it the other way before.

web/Services/PostcodeService.cs
cec200wire: I was wondering why the () kept on showing up as not necessary!

web/Services/PostcodeService.cs
cec200wire: I'd have to agree with you there - I think the name was left-over from an earlier version of the code, but looking at it now it obviously doesn't make sense.

web/Database/TenantDbModel.cs
```diff
@@ -11,7 +12,8 @@ public class TenantDbModel
     public string? Phone { get; set; }
     
     public string? Postcode { get; set; }
-    
+    public decimal? Lat { get; set; }
+    public decimal? Lon { get; set; }
```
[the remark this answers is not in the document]
cec200wire: The tenant isn't required to provide a postcode currently, so there's no guarantee that Lat and Lon will exist (if Postcode is null, Lat and Lon will both also be null).

web/Views/Admin/TenantList.cshtml
cec200wire: I'll be honest - I copied this button from another page, I don't even know if it needs spacing out.

web/Controllers/AdminController.cs
cec200wire: It could be merged into the final step (i.e. when the import is known to be succeeding), but I though it was clearer to separate it out.

web/Services/PostcodeService.cs
cec200wire: BulkAdd, regardless of list length, will use the Postcodes.io Bulk Look-up API, whereas this uses the single Look-up API. In theory, a single postcode could just be dealt with as a 'Bulk' look-up of 1 (the API is fine with this) - so as long as this isn't a problem for reasons of efficiency/cost/anything else, everything can be sent through the Bulk methods.

web/Services/PostcodeService.cs
cec200wire: Would the state of postcodesToLookup be saved at the moment of the Task being added to the list, or would the fact that it immediately gets set to empty cause any problems storing a list of Tasks to await later?

web/Controllers/AdminController.cs
MaxTheComputerer: I don't mind either way tbh, it just felt like it was part of the overall import operation seeing as it updates the tenant table

web/Services/PostcodeService.cs
```diff
@@ -0,0 +1,116 @@
...
+            return postcodeResponseList;
+        }
+        var jObjectResponse = JObject.Parse(bulkResponseBody);
+        var postcodeList = jObjectResponse["result"]!.ToList();
+        foreach (var postcode in postcodeList)
+        {
+            var postcodeDbModel = postcode.ToObject<PostcodeioResponseModel>();
+            postcodeResponseList.Add(postcodeDbModel!);
+        }
+        return postcodeResponseList;
```
oskarwilliams378, line 102: We can actually simplify this down a fair amount :). If you go through and make all of the method argument and return types IEnumerable instead of Lists, then this is what we get
```suggestion
    {
        if (string.IsNullOrEmpty(bulkResponseBody))
        {
            return new List<PostcodeioResponseModel>();
        }
        var jObjectResponse = JObject.Parse(bulkResponseBody);
        var postcodeList = jObjectResponse["result"]!;
        return postcodeList.Select(postcode => postcode.ToObject<PostcodeioResponseModel>()!);

```

In general it is better to use IEnumerable as often as possible instead of List, this is because as we are going through building our enumerable we are basically creating an order of operations, so starting with a, then perform method 1, then perform method 2, etc. When we call ToList it runs through and iterates and performs all these methods in actuality. The reason we want to call this ToList as late as possible, is that the black box of C#, when it compiles, it can simplify and speed up all the different methods we are asking it to perform on an enumerable before making it a list. This means that we can have a situation where instead of doing steps 1 and 2, then going to list, then doing steps 3 and 4 and doing to list, c# can merge steps 2 and 3 into one quicker step 5, and do steps 1 5 and 4 and then do a ToList.

TLDR, use IEnumerable and avoid ToList wherever possible (but don't worry about it too much as its a performance thing and this is not going to be a service that suffers as a result of very slightly less performant code)

web/Services/PropertyService.cs
```diff
@@ -219,9 +219,15 @@ public string CreateNewPublicViewLink(int propertyId)
...
-        if (model.Result == null || model.Result.Lat == null || model.Result.Lon == null)
+        var postcode = _postcodeService.FormatPostcode(postalCode);
+        var postcodeList = new List<string> { postcode };
+        await _postcodeService.AddPostcodesToDatabaseIfAbsent(postcodeList);
+        if (postcode == "")
+        {
+            return null;
+        }
+        var model = _dbContext.Postcodes.SingleOrDefault(p => p.Postcode == postcode);
+        if (model == null || model.Lat == null || model.Lon == null)
```
oskarwilliams378, line 230: you can combine the first 2 checks with a model?.Lat == null

web/Services/TenantService.cs
```diff
@@ -0,0 +1,53 @@
...
+        _postcodeService = postcodeService;
+    }
+
+    public async Task<List<TenantDbModel>?> SortTenantsByLocation(string postalCode, int page, int perPage)
+    {
+        var postcode = _postcodeService.FormatPostcode(postalCode);
+        var postcodeList = new List<string> { postcode };
+        await _postcodeService.AddPostcodesToDatabaseIfAbsent(postcodeList);
+        var targetLocation = _dbContext.Postcodes.SingleOrDefault(p => p.Postcode == postcode);
+        if (targetLocation == null || targetLocation.Lat == null || targetLocation.Lon == null)
```
oskarwilliams378, line 28: Same thing about combining these checks :)

oskarwilliams378 (changes requested):

web/Services/PostcodeService.cs
```diff
@@ -0,0 +1,116 @@
...
+    }
+
+    private async Task<List<PostcodeioResponseModel>> BulkGetLatLonForPostcodes(List<string> postcodesToLookup)
+    {
+        var bulkResponseBody = await MakeBulkApiRequestToPostcodeApi(postcodesToLookup);
+        var postcodeResponse = TurnBulkResponseBodyToModel(bulkResponseBody);
+        return postcodeResponse;
+    }
+
+    private async Task<string> MakeBulkApiRequestToPostcodeApi(List<string> postcodes)
```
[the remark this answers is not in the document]
cec200wire: We're getting a response from the API, I'm not really sure what more useful name there could be.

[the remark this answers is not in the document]
oskarwilliams378: Maybe something like GetPostcodeApiResponseForPostcodes? The Make implies that the code will make a request, not that it will return anything

web/Services/PostcodeService.cs
```diff
@@ -0,0 +1,116 @@
...
+        {
+            AddPostcodesToPostcodeDb(postcodeGroups["valid"]);
+        }
+        if (postcodeGroups.ContainsKey("invalid") && postcodeGroups["invalid"].Length > 0)
+        {
+            _logger.LogWarning($"{postcodeGroups["invalid"].Length} postcodes cannot be converted to coordinates.");
+        }
+    }
+
+    private async Task<List<PostcodeioResponseModel>> BulkGetLatLonForPostcodes(List<string> postcodesToLookup)
```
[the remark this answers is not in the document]
cec200wire: PostcodeioResponse currently only consists of the postcode (known on input anyway) and the Lat and Lon, but sure.

web/Controllers/AdminController.cs
cec200wire: We still have to run over the table twice, since we're first adding the (missing) postcode to LatLon conversions to the postcode table and then using that table to add the LatLon data to the tenant table - I originally tried to do it all at once, but it caused database errors (it can't write to the Postcode table in the middle of writing to the Tenant table, or something to that effect).

oskarwilliams378: So there is a way to make this all work nicely, but I will leave that for a later ticket

web/Controllers/AdminController.cs
```diff
@@ -277,17 +277,24 @@ public async Task<ActionResult> ImportTenants(IFormFile csvFile)
...
         
-        if (furtherFlashResponse.FlashMessages.Count() == 0)
+        if (flashResponse.FlashMessages.Count() == 0)
         {
             _logger.LogInformation("Successfuly imported all tenant data.");
             AddFlashMessage("success", "Successfully imported all tenant data.");
         }
+        
+        flashResponse =  _csvImportService.AddLatLonToTenantDb();
+        for (int i = 0; i < flashResponse.FlashMessages.Count(); i++)
```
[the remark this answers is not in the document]
oskarwilliams378: This hasn't been fixed, also there are still a lot of unnecessary () in this file :)

web/Services/PostcodeService.cs
```diff
@@ -0,0 +1,109 @@
...
+    }
+
+    public async Task AddPostcodesToDatabaseIfAbsent(IEnumerable<string> postcodes)
+    {
+        var postcodeResponseList = new List<PostcodeioResponseModel>();
+        var postcodesToLookupChunks = postcodes.Distinct()
+            .Where(p => p != "" && _dbContext.Postcodes.All(dbRecord => dbRecord.Postcode != p))
+            .Chunk(100);
+
+        var addPostcodeChunkToResponseList = postcodesToLookupChunks.Select(async p => postcodeResponseList.AddRange(await GetPostcodeDetails(p.ToList())));
```
oskarwilliams378, line 42: One last ToList that can bbe removed :)

web/Services/PropertyService.cs
```diff
@@ -219,9 +219,15 @@ public string CreateNewPublicViewLink(int propertyId)
...
-        var model = _postcodeApiService.TurnResponseBodyToModel(responseBody);
-        if (model.Result == null || model.Result.Lat == null || model.Result.Lon == null)
+        var postcode = _postcodeService.FormatPostcode(postalCode);
+        var postcodeList = new List<string> { postcode };
+        await _postcodeService.AddPostcodesToDatabaseIfAbsent(postcodeList);
+        if (postcode == "")
+        {
+            return null;
+        }
+        var model = _dbContext.Postcodes.SingleOrDefault(p => p.Postcode == postcode);
```
oskarwilliams378, line 229: I have jsut realised, can we not combine these 2? as in if the postcode is empty string the model will be null

oskarwilliams378 (commented):
A couple of very minor comments, otherwise approved :)

oskarwilliams378 (dismissed):

web/Controllers/AdminController.cs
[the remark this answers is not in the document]
cec200wire: I've tried to get rid of them all now, hopefully they don't all creep back in rebase/merges.

web/Services/PropertyService.cs
cec200wire: I thought that to, but it was causing an 'Object reference not set to an instance of an object error' so I kept both checks in

TomHanmer29 (approved):

PR #143, opened 2022-08-16
BNH-164 & 93 - property deletion

cec200wire:
Two small fixes in the property deletion code:
BNH-164: when a property is deleted, the container on Azure Blob Storage which holds the images for that property is also deleted (to avoid storage of unneeded, and mostly inaccessible, files)
BNH-93: deleting a property which did not exist (e.g .by having the same property open in two tabs and deleting it in one tab, then attempting to in the other) was returning a (browser) 405 error - why it was 405, not 404, is slightly unclear. It now returns the user to the list of their properties (if a landlord), admin dashboard (if non-landlord admin) or home page (if nothing important), with a flash message informing them the property does not exist.

### DIFF
--- a/web/Controllers/PropertyController.cs
+++ b/web/Controllers/PropertyController.cs
@@ -36,20 +36,26 @@ public class PropertyController : AbstractController
         var propDB = _propertyService.GetPropertyByPropertyId(propertyId);
         if (propDB == null)
         {
-            _logger.LogWarning("Property with ID {PropertyId} does not exist", propertyId);
-            return StatusCode(404);
+            _logger.LogWarning($"Property with Id {propertyId} does not exist");
+            AddFlashMessage("danger", $"Property with Id {propertyId} does not exist");
+            if (GetCurrentUser().LandlordId != null)
+            {
+                return RedirectToAction("ViewProperties", "Landlord", new { id = GetCurrentUser().LandlordId });
+            }
+            return RedirectToAction("Index", "Home");
         }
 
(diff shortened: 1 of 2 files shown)


web/Controllers/PropertyController.cs
```diff
@@ -36,22 +36,26 @@ public ActionResult DeleteProperty(int propertyId)
...
+            }
+            return RedirectToAction("Index", "Home");
         }
 
         if (GetCurrentUser().IsAdmin == false && GetCurrentUser().LandlordId != propDB.LandlordId)
         {
-            _logger.LogWarning("You do not have access to any property with ID {PropertyId}.", propertyId);
-            return StatusCode(404);
+            _logger.LogWarning($"User {GetCurrentUser().Id} does not have access to any property with ID {propertyId}.");
+            return StatusCode(405);
```
oskarwilliams378, line 51: I think instead of a 405 this should be a 403, as the current user is not allowed to access this (ie Forbidden) not that the server knows the request method, but the target resource doesn't support this method (ie Method Not Allowed) as that would be saying that the property would be unable to delete itself, not that the landlord was unauthorised in deleting it

oskarwilliams378 (approved):
One very minor change on what the status code should be, other than that looks good and happy to merge :)

web/Controllers/PropertyController.cs
cec200wire: Yes, that's just me mis-remembering the correct code - I meant 403.

SleepyUser (approved):

PR #146, opened 2022-08-18
BNH-99 Profile pictures

MaxTheComputerer:
- Users' profile pictures from Google are stored in the database as URLs
- These are automatically fetched and updated
- If a landlord is linked to a user and the user has a profile picture, then this is displayed on the profile page, otherwise a default is used

Screenshot shows a landlord's profile picture on their profile page, and the current admin's picture in the navbar
image

image


web/Auth/ClaimsTransformer.cs
```diff
@@ -32,14 +32,20 @@ public async Task<ClaimsPrincipal> TransformAsync(ClaimsPrincipal principal)
...
             // If they don't exist yet, create them
-            databaseUser = await CreateUserInDatabase(googleAccountId, existingClaimsIdentity);
-            _logger.LogInformation("Created new user '{UserEmail}' in database with id {UserId}", databaseUser.GoogleEmail, databaseUser.Id);
+            databaseUser = await CreateUserInDatabase(googleAccountId, googlePhotoUrl, existingClaimsIdentity);
+            _logger.LogInformation("Created new user '{UserEmail}' in database with id {UserId}",
+                databaseUser.GoogleEmail, databaseUser.Id);
+        }
+        else
+        {
+            await UpdateUserProfilePhoto(databaseUser, googlePhotoUrl);
```
oskarwilliams378, line 48: Is this in order to update existing users?

web/Views/Admin/AdminDashboard.cshtml
```diff
@@ -1,4 +1,6 @@
-﻿@model AdminDashboardViewModel
+﻿@using BricksAndHearts.Auth
+@using Microsoft.AspNetCore.Mvc.TagHelpers
```
oskarwilliams378, line 2: Do we need this? The _ViewImports is importing this :)

oskarwilliams378 (changes requested):
Very minor comments

web/Auth/ClaimsTransformer.cs
MaxTheComputerer: Yes!

web/Views/Admin/AdminDashboard.cshtml
MaxTheComputerer: Yeah it seems we don't need it, I think Rider automatically imports it

web/Auth/ClaimsTransformer.cs
oskarwilliams378: could you make the else be an else if instead then, checking that if that database user already has that googlePhotoUrl?

MaxTheComputerer: Ah yeah good idea!

oskarwilliams378 (approved):

cec200wire (approved):

PR #147, opened 2022-08-18
BNH-154 - paginate tables

cec200wire:
Principally intended to paginate tables, this also includes a very small bit of styling on the table pages (mostly just aligning a couple of buttons).

Tenant List and Match List methods are currently a bit messy - a refactor of these, in order to combine the filtering and sorting, is in progress on BNH-168 so I've tried to change the minimum possible while also making pagination work. There will be some slightly nasty merge conflicts between this and BNH-168, but we can deal with that when we need to.

Also, on my local version the site currently hits an exception immediately on being opened, due to something to do with Profile images. The bug is very unlikely to be in the new code on this branch, since I don't think I've been near anything which requires a profile picture, so I'm raising this pull request now in order to avoid unnecessary delays (while I attempt to figure out how to fix the problem).


web/Controllers/AdminController.cs
```diff
@@ -133,20 +136,37 @@ public async Task<ActionResult> GetAdminList()
 
     [Authorize(Roles = "Admin")]
     [HttpGet]
-    public async Task<IActionResult> LandlordList(LandlordListModel landlordListModel)
+    public IActionResult LandlordList(bool? isApproved = null, bool? isAssigned = null, int page = 1, int landlordsPerPage = 10)
     {
-        landlordListModel.LandlordList = await _adminService.GetLandlordList(landlordListModel);
-        return View(landlordListModel);
+        var landlords = _adminService.GetLandlordList(isApproved, isAssigned);
```
oskarwilliams378, line 141: so I would definitely suggest adding in the pagination into the service, and then calling that here. That way we can be certain the pagination is called before any ToList, and that the pagination happens in the DB rather than in memory :)

oskarwilliams378: In order for it to work with how this is working, you could have the method take in the pagination parameters as well and then return both the list and the count

web/Services/TenantService.cs
```diff
@@ -44,10 +44,8 @@ public TenantService(BricksAndHeartsDbContext dbContext, IPostcodeService postco
                 FROM dbo.Tenant
                 WHERE Lon is not NULL and Lat is not NULL
                 ORDER BY distance
-                OFFSET {perPage * (page - 1)} ROWS
-                FETCH NEXT {perPage} ROWS ONLY
                 "
             );
-        return await tenants.ToListAsync();
+        return tenants.AsEnumerable();
```
oskarwilliams378, line 49: So we should be using ToList here. A List is ordered, whereas an Enumerable is not, this means that the ordering here is not guaranteed to remain. Its the same in the other services too :)

web/Services/TenantService.cs
```diff
@@ -19,15 +19,15 @@ public TenantService(BricksAndHeartsDbContext dbContext, IPostcodeService postco
...
+    public async Task<IEnumerable<TenantDbModel>> SortTenantsByLocation(string postalCode)
     {
         var postcode = _postcodeService.FormatPostcode(postalCode);
         var postcodeList = new List<string> { postcode };
         await _postcodeService.AddPostcodesToDatabaseIfAbsent(postcodeList);
         var targetLocation = _dbContext.Postcodes.SingleOrDefault(p => p.Postcode == postcode);
         if (targetLocation?.Lat == null || targetLocation.Lon == null)
         {
-            return null;
+            return null!;
```
oskarwilliams378, line 30: Could this not return an empty list instead?

oskarwilliams378 (changes requested):
Some logic changes to fix before I go over the rest of the code

web/Services/TenantService.cs
cec200wire: Yes, I considered that before but making an empty IQueryable/IEnumerable (which would have been required before) was more difficult to figure out - now that everything will return Lists again, I've made it return as an empty list.

oskarwilliams378: So an empty list is an IEnumerable. A List implement IList, which extends ICollection, which extends IEnumerable, so a List implements the interface IEnumerable :)

web/Services/PropertyService.cs
```diff
@@ -172,23 +171,55 @@ public bool IsUserAdminOrCorrectLandlord(BricksAndHeartsUser currentUser, int pr
...
+                  * cos( radians( Lon ) - radians({model.Lon}) )
+                  + sin ( radians({model.Lat}) )
+                  * sin( radians( Lat ) )
+                    )
+                ) AS distance 
+                FROM dbo.Property
+                WHERE Lon is not NULL and Lat is not NULL
+                ORDER BY distance
+                OFFSET {propPerPage * (page - 1)} ROWS
+                FETCH NEXT {propPerPage} ROWS ONLY
```
oskarwilliams378, line 208: if we remove these 2 lines, and make this assign to a properties IQueryable instead of a list (and then remove the Skip and Take in the other cases) we can then just add the skip and take to the end of the file and only have it written once :)

web/Views/Admin/LandlordList.cshtml
```diff
@@ -93,4 +100,24 @@
         </tr>
     }
     </tbody>
-</table>
\ No newline at end of file
+</table>
+
+@{ var from = (Model.Page - 1) * Model.LandlordsPerPage + 1;
```
oskarwilliams378, line 105: could we put new lines after the `@{` :)

web/Views/Admin/PropertyList.cshtml
```diff
@@ -93,18 +107,19 @@
 @{ var from = (Model.Page - 1) * Model.PropPerPage + 1;
     var to = Math.Min((Model.Page - 1) * Model.PropPerPage + Model.PropPerPage, Model.Total);}
 
-<p>Displaying properties @from to @to of @Model.Total</p>
+@if (Model.Properties.Count > 0)
+{<p>Displaying properties @from to @to of @Model.Total</p>}
 
 @if (Model.Page > 1)
```
oskarwilliams378, line 113: Could these next and previous page button be extracted into a component?

web/Views/Admin/PropertyList.cshtml
```diff
@@ -7,42 +7,56 @@
...
-        <div class="col-8">
-            @using (Html.BeginForm("SortPropertiesByLocation", "Property", FormMethod.Get))
-            {
-                <input class="form-control" id="postcode" name="postcode" required placeholder="Enter postcode to sort by location" pattern="([Gg][Ii][Rr] 0[Aa]{2})|((([A-Za-z][0-9]{1,2})|(([A-Za-z][A-Ha-hJ-Yj-y][0-9]{1,2})|(([A-Za-z][0-9][A-Za-z])|([A-Za-z][A-Ha-hJ-Yj-y][0-9][A-Za-z]?))))\s?[0-9][A-Za-z]{2})"/>
-                <button type="submit" class="btn custom-btn-primary"> Sort </button>
-            }
+    @using (Html.BeginForm("PropertyList", "Property", FormMethod.Get))
+    {
+        <div class="row">
+            <select class="form-select" name="sortBy"
```
oskarwilliams378, line 13: Could we use the dropdown list here instead of the select :)

web/Views/Admin/TenantList.cshtml
```diff
@@ -136,6 +147,26 @@
     </tbody>
 </table>
 
+@{ var from = (Model.Page - 1) * Model.TenantsPerPage + 1;
```
oskarwilliams378, line 150: This looks very much like another change, can we extract all this into a component :)

web/Views/Admin/TenantMatchList.cshtml
```diff
@@ -1,52 +1,61 @@
...
+            <td>@tenant.Type</td>
+            <td>@tenant.HasPet</td>
+            <td>@tenant.ETT</td>
+            <td>@tenant.UniversalCredit</td>
+            <td>@tenant.HousingBenefits</td>
+            <td>@tenant.Over35</td>
+            <td>
+                @using (Html.BeginForm("SendMatchLinkEmail", "Admin", new
+                {
+                    propertyLink = Model.CurrentProperty.Description, /*TODO: change this from description to link*/
```
oskarwilliams378, line 49: Do we have a ticket for this?

web/Views/Landlord/Properties.cshtml
```diff
@@ -76,4 +86,24 @@ else
     }
     </tbody>
 </table>
-<a class="btn btn-primary" href="@Url.Action("AddNewProperty_Begin", "Property", new { Model.Owner?.LandlordId })">Add new property</a> @*Add landlord ID*@
\ No newline at end of file
+
+@{ var from = (Model.Page - 1) * Model.PropPerPage + 1;
```
oskarwilliams378, line 90: Yeah definitely make this into a component :)

web/Controllers/LandlordController.cs
```diff
@@ -206,10 +206,12 @@ public async Task<IActionResult> ViewProperties(int? id = null)
         }
 
         var databaseResult = _propertyService.GetPropertiesByLandlord(id.Value);
-        var listOfProperties = databaseResult.Select(PropertyViewModel.FromDbModel).ToList();
+        var listOfProperties = databaseResult.Select(PropertyViewModel.FromDbModel).Skip((page - 1) * propPerPage)
```
oskarwilliams378, line 209: Could we have the pagination in the service please :)

oskarwilliams378 (changes requested):

web/Views/Admin/PropertyList.cshtml
cec200wire: Max is sorting this on his branch - we agreed that there's no point duplicating the work.

web/Views/Admin/PropertyList.cshtml
cec200wire: I've extracted the 'Displaying [items] [min] to [max] of [total]' code, but the next and previous page buttons are awkward (there are different values being passed in depending on the page - sometimes a sortBy, sometimes a filter, sometimes a postcode).

web/Views/Admin/TenantMatchList.cshtml
cec200wire: It's in BNH-162.

web/Controllers/LandlordController.cs
cec200wire: Sure, I forgot there was this page which needed re-paginating (every other paginated page is an Admin View, so easier to keep track of)

web/Services/PropertyService.cs
cec200wire: If I do that, I get this error:
SqlException: The ORDER BY clause is invalid in views, inline functions, derived tables, subqueries, and common table expressions, unless TOP, OFFSET or FOR XML is also specified. Incorrect syntax near 'OFFSET'. Invalid usage of the option NEXT in the FETCH statement.
I'm not very familiar with the SQL commands so I'm not entirely sure what this error is, any idea what it is/how to fix it?

cec200wire: I've managed to fix this by adding OFFSET 0 ROWS - so the skip and take can still appear at the end, but SQL stops having issues.

web/Views/Admin/PropertyList.cshtml
oskarwilliams378: From what it looks like, could the pagination model just take in a previousHref and nextHref? you can still define the links in their relative files using the Url.Action, but wouldn't need all the html etc

oskarwilliams378 (changes requested):
One more comment on the pagination

web/Views/Admin/PropertyList.cshtml
cec200wire: Yes, that would work I suppose.

oskarwilliams378 (approved):

PR #148, opened 2022-08-18
BNH-141 Style property tables

MaxTheComputerer:
- Admin and landlord property lists now share a partial for the table itself
- Table styled to match designs
- On mobile, table is scrollable
- To make the wide tables fit, we change from a `container` to a `container-fluid` for pages which have `TempData["Wide"]` set to `true`

Landlord:
image

Admin:
image

Landlord mobile:
image

### DIFF
--- a/web/Views/Admin/PropertyList.cshtml
+++ b/web/Views/Admin/PropertyList.cshtml
@@ -1,113 +1,56 @@
 ﻿@model PropertiesDashboardViewModel
 
 @{
-    ViewData["Title"] = "All Properties List";
+    ViewData["Title"] = "Properties";
 }
 
-<h2>All Properties List</h2>
+<div class="mx-md-5">
+    <h1 class="font-h1 mt-5">Properties</h1>
 
-<div class="mb-3">
-    <div class="row">
-        <div class="col-4">
-            @using (Html.BeginForm("SortProperties", "Property", FormMethod.Get))
-            {
-                <select class="form-select" id="sortBy" name="sortBy">
+    <div class="row mt-4">
+        <div class="col-auto me-auto mb-2">
+            <form asp-action="SortProperties" asp-controller="Property" method="get" class="row mx-0">
+                <select class="form-select col h-auto me-3" id="sortBy" name="sortBy">
                     <option value="Availability">Sort by availability</option>
                     <option value="Rent">Sort by rent</option>
                 </select>
-                <button type="submit" class="btn custom-btn-primary"> Display results </button>
-            }
+                <button type="submit" class="btn btn-outline-primary col">Sort</button>
+            </form>
         </div>
-        <div class="col-8">
-            @using (Html.BeginForm("SortPropertiesByLocation", "Property", FormMethod.Get))
-            {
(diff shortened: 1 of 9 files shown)


web/Views/Shared/_Layout.cshtml
```diff
@@ -85,7 +85,7 @@
 
 
 <main role="main">
-    <div class="container mt-3 pb-4">
+    <div class="container@(TempData["Wide"] is true ? "-fluid" : "") mt-3 pb-4">
```
oskarwilliams378, line 88: I think possibly we should change this from just `"Wide"`. Is there a more descriptive name we could use?

web/Views/Admin/PropertyList.cshtml
```diff
@@ -1,113 +1,56 @@
...
-<div class="mb-3">
-    <div class="row">
-        <div class="col-4">
-            @using (Html.BeginForm("SortProperties", "Property", FormMethod.Get))
-            {
-                <select class="form-select" id="sortBy" name="sortBy">
+    <div class="row mt-4">
+        <div class="col-auto me-auto mb-2">
+            <form asp-action="SortProperties" asp-controller="Property" method="get" class="row mx-0">
+                <select class="form-select col h-auto me-3" id="sortBy" name="sortBy">
```
oskarwilliams378, line 13: This won't stay filled when we send this form and get back to this page, could we possibly make this a DropdownListFor like in the Landlord list :)

web/Views/Landlord/Properties.cshtml
```diff
@@ -1,79 +1,33 @@
...
+            @("My properties")
+        }
+        else
+        {
+            @(Model.Owner.FirstName + "'s properties")
+        }
+    </h1>
+
+    <div class="row mt-4">
+        @* Sorting and filtering goes here *@
```
oskarwilliams378, line 25: Does this have a ticket number? :)

web/Views/Property/_PropertyTable.cshtml
```diff
@@ -0,0 +1,97 @@
...
+        <tbody>
+        @foreach (var prop in Model.Properties)
+        {
+            <tr>
+                <td class="text-nowrap">
+                    <partial name="../Property/Details/_AvailabilityPill" model="prop"/>
+                </td>
+                <td>
+                    @prop.Address.AddressLine1,
+                    @prop.Address.AddressLine2@(prop.Address.AddressLine2 != null ? "," : "")
```
oskarwilliams378, line 39: I think it would be nice to combine these 2, and below for addressline 3 as well :) 
```suggestion
                    @(prop.Address.AddressLine2 == null ? "" : prop.Address.AddressLine2 + ",")

```

oskarwilliams378 (changes requested):

web/Views/Shared/_Layout.cshtml
MaxTheComputerer: Changed to `"FullWidthPage"`

web/Views/Landlord/Properties.cshtml
MaxTheComputerer: I don't think there's a ticket for it at the moment, I just left a space for it in case someone adds in the future.

oskarwilliams378 (approved):

PR #150, opened 2022-08-19
BNH-146 - warn if admin tries to become landlord

cec200wire:
If an Admin tries to sign themselves up as a landlord, the landlord registration page has a warning flash message:
image

### DIFF
--- a/web/Views/Shared/_Layout.cshtml
+++ b/web/Views/Shared/_Layout.cshtml
@@ -86,11 +86,13 @@
 
 <main role="main">
     <div class="container mt-3 pb-4">
-        @if (TempData["FlashMessages"] is string[] flashMessages && TempData["FlashTypes"] is string[] flashTypes)
+        @if (TempData["FlashMessages"] != null && TempData["FlashTypes"] != null)
         {
-            for (var i = 0; i < flashMessages.Length; i++)
+            var flashMessages = TempData["FlashMessages"] as List<string>;
+            var flashTypes = TempData["FlashTypes"] as List<string>;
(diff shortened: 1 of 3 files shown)


web/Views/Shared/_Layout.cshtml
```diff
@@ -86,11 +86,13 @@
...
         {
-            for (var i = 0; i < flashMessages.Length; i++)
+            var flashMessages = TempData["FlashMessages"] as List<string>;
+            var flashTypes = TempData["FlashTypes"] as List<string>;
+            for (var i = 0; i < flashMessages!.Count; i++)
             {
-                <div class="alert alert-@flashTypes[i]" role="alert">
+                <div class="alert alert-@flashTypes![i]" role="alert">
                     @flashMessages[i]
                 </div>
```
oskarwilliams378, line 97: Why has this changed back?

oskarwilliams378 (changes requested):

web/Views/Shared/_Layout.cshtml
cec200wire: The way it was changed to, any new flash messages weren't displaying - the ones which were in the code previous to that change would still appear, but every time AddFlashMessage was added somewhere, the message would be added to TempData but would not be registered here by the as string[] and so wouldn't display.

oskarwilliams378: Hmm ok, sounds very very odd but ok

oskarwilliams378 (approved):

PR #153, opened 2022-08-19
BNH-NONE: Fix postcode linking and replace lat/lon with Location, allowing for better sql queries

oskarwilliams378:
Replace the postcode and lat/lon in properties and tenants with an actual foreign key to the Postcodes DB. This will then allow for much easier distance calculation, allow less data in the properties and tenants table, and have one source of truth for postcodes


web/Database/PropertyDbModel.cs
```diff
@@ -20,9 +20,8 @@ public class PropertyDbModel
     public string? AddressLine3 { get; set; }
     public string? TownOrCity { get; set; }
     public string? County { get; set; }
-    public string Postcode { get; set; } = string.Empty;
-    public decimal? Lat { get; set; } // latitude
-    public decimal? Lon { get; set; } // longitude
+    public PostcodeDbModel? Postcode { get; set; }
+    public string? PostcodeId { get; set; }
```
emilypatsko, line 24: Pedantic but I feel like these lines should be the other way round, like `LandlordId` and `Landlord` above 😁

web/Services/PostcodeService.cs
```diff
@@ -41,15 +64,16 @@ public async Task AddPostcodesToDatabaseIfAbsent(IEnumerable<string> postcodes)
 
         var addPostcodeChunkToResponseList = postcodesToLookupChunks.Select(async p => postcodeResponseList.AddRange(await GetPostcodeDetails(p)));
         await Task.WhenAll(addPostcodeChunkToResponseList);
-        
+
         var postcodeGroups = postcodeResponseList
-                .GroupBy(p => p.Result?.Lat == null || p.Result.Lon == null || p.Result.Postcode == null)
-                .ToDictionary(grouping => grouping.Key ? "invalid" : "valid", grouping => grouping.ToArray());
-    
+            .GroupBy(p => p.Result?.Lat == null || p.Result.Lon == null || p.Result.Postcode == null)
```
emilypatsko, line 69: Github seems to think these `null` checks will always be false as these are `double` not `double?`

web/Database/PropertyDbModel.cs
oskarwilliams378: SUre :)

emilypatsko (approved):

emilypatsko (approved):

PR #154, opened 2022-08-19
BNH-151 - style Admin Dashboard

cec200wire:
The Admin Dashboard now looks like this:
image

Compared to the Figma design of this:
image

(I discussed changing the 'Admin' stats with 'Tenant' stats with Shiv, as it seemed a bit illogical for Admins to have information on each other).

For any logged-in non-admin user, the page will appear like this:
image


web/Services/AdminService.cs
```diff
@@ -47,6 +48,14 @@ public AdminService(BricksAndHeartsDbContext dbContext, ILogger<AdminService> lo
         UserDbModel userFromId = _dbContext.Users.SingleOrDefault(u => u.Id == userId)!;
         return userFromId;
     }
+    
+    public AdminCountModel CountAdmins()
+    {
+        AdminCountModel adminCounts = new AdminCountModel();
+        adminCounts.CurrentAdmins = _dbContext.Users.Count(u => u.IsAdmin == true);
+        adminCounts.PendingAdmins = _dbContext.Users.Count(u => u.HasRequestedAdmin == true && u.IsAdmin == false);
+        return adminCounts;
```
oskarwilliams378, line 57: this could just be 
```suggestion
        return new AdminCountModel
        {
            CurrentAdmins = _dbContext.Users.Count(u => u.IsAdmin == true),
            PendingAdmins = _dbContext.Users.Count(u => u.HasRequestedAdmin == true && u.IsAdmin == false)
        };

```

web/Services/TenantService.cs
```diff
@@ -18,6 +20,14 @@ public TenantService(BricksAndHeartsDbContext dbContext, IPostcodeService postco
         _dbContext = dbContext;
         _postcodeService = postcodeService;
     }
+    
+    public TenantCountModel CountTenants()
+    {
+        TenantCountModel tenantCounts = new TenantCountModel();
+        tenantCounts.RegisteredTenants = _dbContext.Tenants.Count();
+        tenantCounts.LocatedTenants = _dbContext.Tenants.Count(t => t.Postcode != null);
+        return tenantCounts;
```
oskarwilliams378, line 29: Equally this can be replaced with 
```suggestion
        return new TenantCountModel
        {
            RegisteredTenants = _dbContext.Tenants.Count(),
            LocatedTenants = _dbContext.Tenants.Count(t => t.Postcode != null)
        };

```

oskarwilliams378 (changes requested):

PR #155, opened 2022-08-22
BNH-162 Public property page

MaxTheComputerer:
- Refactored public link generation to occur when a property is created
- The tokens are also generated on a couple of extra pages for any existing properties which don't have one
- The link is sent in the tenant match email, and is copy-able from a property's details page
- The endpoint of this link has been styled to match the designs, reusing components from the details page

image
image
image
image


web/Services/PropertyService.cs
```diff
@@ -103,15 +109,17 @@ public void UpdateProperty(int propertyId, PropertyViewModel updateModel, bool i
...
         dbModel.AcceptsCouple = updateModel.LandlordRequirements.AcceptsCouple ?? dbModel.AcceptsCouple;
         dbModel.AcceptsFamily = updateModel.LandlordRequirements.AcceptsFamily ?? dbModel.AcceptsFamily;
         dbModel.AcceptsPets = updateModel.LandlordRequirements.AcceptsPets ?? dbModel.AcceptsPets;
         dbModel.AcceptsCredit = updateModel.LandlordRequirements.AcceptsCredit ?? dbModel.AcceptsCredit;
         dbModel.AcceptsBenefits = updateModel.LandlordRequirements.AcceptsBenefits ?? dbModel.AcceptsBenefits;
         dbModel.AcceptsNotEET = updateModel.LandlordRequirements.AcceptsNotEET ?? dbModel.AcceptsNotEET;
         dbModel.AcceptsOver35 = updateModel.LandlordRequirements.AcceptsOver35 ?? dbModel.AcceptsOver35;
-        dbModel.AcceptsWithoutGuarantor = updateModel.LandlordRequirements.AcceptsWithoutGuarantor ?? dbModel.AcceptsWithoutGuarantor;
+        dbModel.AcceptsWithoutGuarantor = updateModel.LandlordRequirements.AcceptsWithoutGuarantor ??
+                                          dbModel.AcceptsWithoutGuarantor;
```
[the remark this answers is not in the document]
MaxTheComputerer: Yeah the IDE's reformatting is a little aggressive!

web/Services/PropertyService.cs
```diff
@@ -165,7 +170,14 @@ public bool IsUserAdminOrCorrectLandlord(BricksAndHeartsUser currentUser, int pr
         return propertyLandlordId == userLandlordId;
     }
 
-    public async Task<(List<PropertyDbModel> PropertyList, int Count)> GetPropertyList(string? sortBy, string? target, int page, int propPerPage)
+    public LandlordDbModel GetPropertyOwner(int propertyId)
+    {
+        return _dbContext.Properties.Include(p => p.Landlord).ThenInclude(l => l.Postcode)
```
oskarwilliams378, line 175: So I meant do 
```suggestion
        return _dbContext.Properties.Include(p => p.Landlord).Include(p => p.Postcode)

```

The Landlord postcode is still just a string as currently we aren't doing anything with it

web/ViewModels/PropertyViewModel.cs
```diff
@@ -12,6 +12,7 @@ public class PropertyViewModel : IValidatableObject
     public int PropertyId;
     public int LandlordId { get; set; }
     public DateTime? CreationTime { get; set; }
+    public string? PublicViewLink { get; set; }
```
oskarwilliams378, line 17: double new line

oskarwilliams378 (changes requested):

web/Services/PropertyService.cs
MaxTheComputerer: Ah I see! This method only returns the Landlord though, is there a reason to include the property's postcode as well?

oskarwilliams378: Yep you're totally right, ignore me

oskarwilliams378 (approved):

PR #157, opened 2022-08-22
BNH-155 - style Contact page

cec200wire:
Contact Us page now looks like this:
image

And on mobile, roughly like this:
image

Compared to a Figma design of this:
image

I have contacted Shiv about whether the social media icons should all be in the website's colour (the primary blue they are currently in) or whether they should be in their traditional colours (i.e. YouTube in red, Twitter in light blue); since she doesn't work on Mondays, it will be a while before I here back, but if it turns out these icons need recolouring it can always be done on a separate branch.


web/Views/Home/ContactUs.cshtml
```diff
@@ -1,43 +1,111 @@
...
+                    </div>
+                </div>
+            </div>
+        </div>
+        <div class="col-md-9">
+            <h1>FAQ</h1>
+            <div class="card custom-card mt-2">
+                <div class="card-body">
+                    <div class="collapse" id="faqCollapse">
+                        <strong>As a tenant, what references do I need to provide to move into a property?</strong>
```
oskarwilliams378, line 85: Could this be turned into a component

oskarwilliams378 (changes requested):

web/Views/Home/_FAQPartial.cshtml
```diff
@@ -0,0 +1,27 @@
+﻿<div class="card custom-card mt-2">
+    <div class="card-body">
+        <div class="collapse" id="faqCollapse">
+            <strong>As a tenant, what references do I need to provide to move into a property?</strong>
+            <p>References will depend on the Landlord’s criteria. These references can be emailed to the Landlord with the support of a Change Ahead volunteer.</p>
```
oskarwilliams378, line 5: I more meant that each <strong> and <p> combination could be a component, which we then pass in a single (string question, string answer). So we could just turn this into a foreach of a List<(string question, string answer)>, where we define the list in the view model

oskarwilliams378 (changes requested):

web/Views/Home/_FAQPartial.cshtml
cec200wire: That makes a lot more sense, I don't know how I misinterpreted that!

oskarwilliams378 (approved):

PR #158, opened 2022-08-22
BNH-138 Style landlord registration form

MaxTheComputerer:
Just front-end tweaks to the landlord registration form to match the designs

image
image
image
image
image

### DIFF
--- a/web/Views/Landlord/_RegisterPagePartial.cshtml
+++ b/web/Views/Landlord/_RegisterPagePartial.cshtml
@@ -1,6 +1,20 @@
 @model LandlordProfileModel
-<div class="registerTab col-md-5" id="Name">
-    <div class="form-group my-3">
+
+@functions
+{
+    void GenerateReviewTableRow(string id, string name)
+    {
+        <tr>
+            <th class="w-25">@name</th>
+            <td class="property-description" id="@(id)Review"></td>
+        </tr>
+    }
+}
+
+<div class="registerTab" id="Name">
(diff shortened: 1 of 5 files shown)


web/Views/Landlord/_RegisterPagePartial.cshtml
```diff
@@ -13,121 +27,173 @@
...
                     <option value="Private Residential Sector">Private Residential Sector</option>
                     <option value="Registered Provider">Registered Provider</option>
                     <option value="Local Authority">Local Authority</option>
                 </select>
                 <span asp-validation-for="LandlordType" class="text-danger"></span>
             </div>
-            <div class="form-group my-3">
+            <div class="my-3">
                 <label asp-for="IsLandlordForProfit" class="form-label"></label>
                 <select asp-for="IsLandlordForProfit" class="form-select" required>
```
oskarwilliams378, line 118: Should we use the Html DropdownListFors or are we not changing any functionality here?

oskarwilliams378 (approved):

web/Views/Landlord/_RegisterPagePartial.cshtml
MaxTheComputerer: I was focusing on just styling the page, but looking at it I think the tag helpers (with the `asp-for` attributes) mean that these dropdowns remain filled in if validation fails. Unless there's another reason to use `DropdownListFor` over tag helpers?

PR #167, opened 2022-08-22
BNH-181 feat: Form buttons now say "submit" if clicking it would not lead to another form page.

SleepyUser:
Buttons now say "submit" if clicking it would not lead to another form page.


cec200wire (commented):
Looks fine to me

oskarwilliams378 (approved):

PR #168, opened 2022-08-22
BNH-147 - disable landlords

cec200wire:
[DATABASE UPDATES IN THIS BRANCH}
Admins can disable landlords (and then re-enable them) through the landlord's profile.
Disabled landlords will see the following display instead of their dashboard:
image
Disabled landlords will be blocked from all other landlord pages, and will not see the landlord nav-bar.

### DIFF
--- a/BricksAndHearts.UnitTests/ControllerTests/Landlord/LandlordControllerTests.cs
+++ b/BricksAndHearts.UnitTests/ControllerTests/Landlord/LandlordControllerTests.cs
@@ -1,6 +1,7 @@
 using System.Linq;
 using System.Threading.Tasks;
 using BricksAndHearts.Database;
+using BricksAndHearts.Enums;
 using BricksAndHearts.Services;
 using BricksAndHearts.ViewModels;
 using FakeItEasy;
@@ -39,7 +40,7 @@ public class LandlordControllerTests : LandlordControllerTestsBase
         formResultModel.Address.AddressLine1 = "Test Road";
 
         A.CallTo(() => LandlordService.RegisterLandlord(formResultModel, unregisteredUser))
-            .Returns((ILandlordService.LandlordRegistrationResult.Success, returnedLandlord));
+            .Returns((LandlordRegistrationResult.Success, returnedLandlord));
 
         // Act
         var result = await UnderTest.RegisterPost(formResultModel) as RedirectToActionResult;
@@ -60,7 +61,7 @@ public class LandlordControllerTests : LandlordControllerTestsBase
         var landlordUser = CreateLandlordUser();
         var landlordProfile = new LandlordProfileModel { LandlordId = landlordUser.Id, MembershipId = "abc" };
         A.CallTo(() => LandlordService.ApproveLandlord(landlordUser.Id, adminUser, "abc"))
-            .Returns(ILandlordService.ApproveLandlordResult.Success);
+            .Returns(ApproveLandlordResult.Success);
 
         // Act
         await UnderTest.ApproveCharter(landlordProfile);
@@ -98,7 +99,7 @@ public class LandlordControllerTests : LandlordControllerTestsBase
         var landlordUser = CreateLandlordUser();
         var landlordProfile = new LandlordProfileModel { LandlordId = landlordUser.Id, MembershipId = "abc" };
         A.CallTo(() => LandlordService.DisableOrEnableLandlord(landlordUser.Id, action))
-            .Returns(ILandlordService.DisableOrEnableLandlordResult.Success);
+            .Returns(DisableOrEnableLandlordResult.Success);
 
         // Act
         var result = await UnderTest.DisableOrEnableLandlord(landlordProfile, action);
@@ -120,7 +121,7 @@ public class LandlordControllerTests : LandlordControllerTestsBase
         var landlordUser = CreateLandlordUser();
         var landlordProfile = new LandlordProfileModel { LandlordId = landlordUser.Id, MembershipId = "abc" };
         A.CallTo(() => LandlordService.DisableOrEnableLandlord(landlordUser.Id, action))
-            .Returns(ILandlordService.DisableOrEnableLandlordResult.ErrorLandlordNotFound);
+            .Returns(DisableOrEnableLandlordResult.ErrorLandlordNotFound);
 
         // Act
         var result = await UnderTest.DisableOrEnableLandlord(landlordProfile, action);
@@ -142,7 +143,7 @@ public class LandlordControllerTests : LandlordControllerTestsBase
         var landlordUser = CreateLandlordUser();
         var landlordProfile = new LandlordProfileModel { LandlordId = landlordUser.Id, MembershipId = "abc" };
         A.CallTo(() => LandlordService.DisableOrEnableLandlord(landlordUser.Id, action))
-            .Returns(ILandlordService.DisableOrEnableLandlordResult.ErrorAlreadyInState);
+            .Returns(DisableOrEnableLandlordResult.ErrorAlreadyInState);
 
         // Act
         var result = await UnderTest.DisableOrEnableLandlord(landlordProfile, action);
@@ -208,7 +209,7 @@ public class LandlordControllerTests : LandlordControllerTestsBase
         A.CallTo(() => LandlordService.CheckForDuplicateEmail(landlordProfileModel)).Returns(false);
         A.CallTo(() => LandlordService.CheckForDuplicateMembershipId(landlordProfileModel)).Returns(false);
         A.CallTo(() => LandlordService.EditLandlordDetails(landlordProfileModel))
-            .Returns(ILandlordService.LandlordRegistrationResult.Success);
+            .Returns(LandlordRegistrationResult.Success);
 
         // Act
         var result = await UnderTest.EditProfileUpdate(landlordProfileModel);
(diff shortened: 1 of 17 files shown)


web/Controllers/LandlordController.cs
```diff
@@ -116,6 +116,7 @@ public async Task<ActionResult> RegisterPost([FromForm] LandlordProfileModel cre
         }
     }
 
+    [Authorize(Roles = "Admin,Landlord")]
```
emilypatsko, line 119:
```suggestion
    [Authorize(Roles = "Admin, Landlord")]
```

web/Controllers/LandlordController.cs
```diff
@@ -136,6 +137,7 @@ public async Task<ActionResult> Profile([FromRoute] int id)
         return View("Profile", viewModel);
     }
 
+    [Authorize(Roles = "Admin,Landlord")]
```
emilypatsko, line 140:
```suggestion
    [Authorize(Roles = "Admin, Landlord")]
```

web/Controllers/LandlordController.cs
```diff
@@ -193,7 +195,40 @@ public async Task<ActionResult> ApproveCharter(LandlordProfileModel landlord)
         AddFlashMessage(flashMessageType, flashMessageBody);
         return RedirectToAction("Profile", "Landlord", new { Id = landlord.LandlordId.Value });
     }
+    
+    [Authorize(Roles = "Admin")]
+    [HttpPost("{id:int}/{action}")]
+    public async Task<ActionResult> DisOrEnableLandlord(LandlordProfileModel landlord, string action)
```
emilypatsko, line 201: Could this be called `DisableOrEnableLandlord` instead? It's a bit clearer

web/Controllers/LandlordController.cs
```diff
@@ -193,7 +195,40 @@ public async Task<ActionResult> ApproveCharter(LandlordProfileModel landlord)
...
+                flashMessageType = "warning";
+                break;
+        }
+
+        _logger.LogInformation(flashMessageBody);
+        AddFlashMessage(flashMessageType, flashMessageBody);
+        return RedirectToAction("Profile", "Landlord", new { Id = landlord.LandlordId.Value });
+    }
 
+    [Authorize(Roles = "Admin,Landlord")]
```
emilypatsko, line 231:
```suggestion
    [Authorize(Roles = "Admin, Landlord")]
```

web/Controllers/LandlordController.cs
```diff
@@ -221,6 +256,7 @@ public async Task<IActionResult> ViewProperties(int? id = null, int page = 1, in
             new PropertyListModel(listOfProperties, properties.Count, landlordProfile, page));
     }
 
+    [Authorize(Roles = "Admin,Landlord")]
```
emilypatsko, line 259: same again, and a few below (not going to comment on all of them)

web/Services/LandlordService.cs
```diff
@@ -231,6 +239,36 @@ public async Task UnapproveLandlord(int landlordId)
...
+                return ILandlordService.DisableLandlordResult.ErrorAlreadyDisabled;
+            }
+
+            landlord.Disabled = true;
+        }
+        else
+        {
+            if (!landlord.Disabled)
+            {
+                return ILandlordService.DisableLandlordResult.ErrorAlreadyDisabled;
```
emilypatsko, line 263: This error seems unintuitive to me - this is getting returned if someone tries to enable an already enabled landlord, right? So I wouldn't expect to see a line of code returning an error that says 'already _dis_ abled'

web/Views/Landlord/Profile.cshtml
```diff
@@ -142,15 +142,32 @@ else
...
+        }
+    }
+   else
+    {
+        using (Html.BeginForm("DisOrEnableLandlord", "Landlord", new { landlordId = Model.LandlordId}, FormMethod.Post))
+        {
+            <input type="hidden" asp-for="Action" value="disable">
+            <button type="submit" class="btn btn-primary"><i class="bi bi-link"></i>Disable landlord</button>
+        }
+    }
```
emilypatsko, line 168: This could be reduced by using ternaries for the input value and the button text, e.g.
`<input type="hidden" asp-for="Action" value=@(Model.Disabled ? "enable" : "disable")>`

emilypatsko (changes requested):
Have left a few comments but haven't looked at the tests yet, will do tomorrow

cec200wire:
I've realised I missed a small part of this ticket (that properties owned by disabled landlords should not appear on the main list of properties) - will go and add that when I have time.

web/Controllers/LandlordController.cs
cec200wire: Ok, I'll try to find all of them.

web/Services/LandlordService.cs
cec200wire: Because the two methods are just the reverse of each other, they're sharing error statuses (with the message returned just substituting in 'disable' or 'enable' - I'll rename it to make it clearer.

web/Views/Landlord/Profile.cshtml
cec200wire: I always forget that ternaries exist!

BricksAndHearts.UnitTests/ControllerTests/Landlord/LandlordControllerTests.cs
```diff
@@ -87,6 +87,25 @@ public async void ApproveCharter_WithoutMembershipId_DisplaysErrorMessage()
         FlashMessages.Should().Contain("Membership ID is required.");
     }
 
+    [Fact]
```
emilypatsko, line 90: Could we turn this test into a `Theory` and have four different test cases (one for each of the possible results returned by the call to `LandlordService`)?
Should also check the result of the call to the controller is the correct thing, i.e. `RedirectToActionResult`

web/Services/LandlordService.cs
```diff
@@ -231,6 +239,36 @@ public async Task UnapproveLandlord(int landlordId)
...
+            return ILandlordService.DisableLandlordResult.ErrorLandlordNotFound;
+        }
+
+        if (action == "disable")
+        {
+            if (landlord.Disabled)
+            {
+                return ILandlordService.DisableLandlordResult.ErrorAlreadyInState;
+            }
+
```
emilypatsko, line 256: remove line break

emilypatsko (changes requested):

BricksAndHearts.UnitTests/ControllerTests/Landlord/LandlordControllerTests.cs
cec200wire: I think I've figured it out but I'm not certain, I'll push what I have.

web/Auth/ClaimsTransformer.cs
```diff
@@ -51,8 +51,11 @@ public async Task<ClaimsPrincipal> TransformAsync(ClaimsPrincipal principal)
         var claims = new List<Claim>(existingClaimsIdentity.Claims);
         // If they're an admin in the database, add the role to their claims
         if (databaseUser.IsAdmin) claims.Add(new Claim(ClaimTypes.Role, "Admin"));
-        if (databaseUser.LandlordId != null) claims.Add(new Claim(ClaimTypes.Role, "Landlord"));
-
+        if (databaseUser.LandlordId != null)
+        {
+            var userLandlord = _context.Landlords.FirstOrDefault(l => l.Id == databaseUser.LandlordId);
```
oskarwilliams378, line 56: So instead of this,we could just grab the landlord at the first call, rather than call again. On line 38 we can add a `.Include(u => u.Landlord)`. Also is there a reason we are using first rather than Single?

If we do it this way then we can make the check into
```
        if (databaseUser.Landlord is { Disabled: false })
        {
            claims.Add(new Claim(ClaimTypes.Role, "Landlord"));
        }
```

web/Services/LandlordService.cs
```diff
@@ -30,6 +30,13 @@ public enum ApproveLandlordResult
         ErrorDuplicateMembershipId,
         Success
     }
+    
+    public enum DisableOrEnableLandlordResult
```
oskarwilliams378, line 34: I would rather if these enums didn't exist on the interface. I think they should either exist on the class implementation of the interface, or in a separate folder :)

web/Services/LandlordService.cs
```diff
@@ -231,6 +239,35 @@ public async Task UnapproveLandlord(int landlordId)
...
+            landlord.Disabled = true;
+        }
+        else
+        {
+            if (!landlord.Disabled)
+            {
+                return ILandlordService.DisableOrEnableLandlordResult.ErrorAlreadyInState;
+            }
+            landlord.Disabled = false;
+        }
```
oskarwilliams378, line 265: 2 things:
1. I think we shouldn't be accepting anything other than disable as enable
2. We can shorten this slightly

So the 2 together would be
```suggestion
        if ((action == "disable" && landlord.Disabled)
            || (action == "enable" && !landlord.Disabled))
        {
            return ILandlordService.DisableOrEnableLandlordResult.ErrorAlreadyInState;
        }

        landlord.Disabled = action switch
        {
            "disable" => true,
            "enable" => false,
            _ => landlord.Disabled
        };

```

oskarwilliams378 (changes requested):

web/Auth/ClaimsTransformer.cs
cec200wire: Ah, I tried grabbing the landlord at first call, but without the .Include(u=> u.Landlord) it was (inexplicably, as far as I could tell) returning null - it's much nicer getting the Landlord directly.
As to why it's First rather than Single - there might have been a reason I did that, or it might have just been that I was tired at the time and not thinking properly - I'll change it.

oskarwilliams378 (approved):

emilypatsko (approved):

PR #169, opened 2022-08-22
BNH-153 Style landlord list

MaxTheComputerer:
Applied styling to match designs

image
image
image

### DIFF
--- a/web/Views/Admin/LandlordList.cshtml
+++ b/web/Views/Admin/LandlordList.cshtml
@@ -1,23 +1,7 @@
-﻿@using Microsoft.Extensions.Primitives
-@model LandlordListModel
+﻿@model LandlordListModel
 
 @{
     ViewData["Title"] = "Landlord List";
-}
-
-@functions
-{
-    string GetCollapseClass()
-    {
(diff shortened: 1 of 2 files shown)


web/Views/Admin/LandlordList.cshtml
```diff
@@ -1,113 +1,80 @@
...
-<a class="btn btn-primary" asp-controller="Landlord" asp-action="RegisterGet" asp-route-createUnassigned="true">Create Unassigned Landlord</a> <br/>
-
-<p></p>
-@using (Html.BeginForm("LandlordList", "Admin", FormMethod.Get))
+@functions
 {
-    <div class="row g-3">
+    string GetCollapseClass()
+    {
+        var isApproved = Context.Request.Query["IsApproved"];
```
oskarwilliams378, line 12: These properties exist on the view model, is there a reason why we are not just using that?

oskarwilliams378 (changes requested):

web/Views/Admin/LandlordList.cshtml
MaxTheComputerer: Ah I didn't realise they were there, that simplifies things a lot!

PR #170, opened 2022-08-23
Bnh 163 match page styling

TomHanmer29:
Styled match page to look like shiv's designs
image
image
image


web/Views/Admin/_MatchListTable.cshtml
```diff
@@ -0,0 +1,82 @@
...
+        return preference == true ? "" : "font-orange";
+    }
+
+    static string GetPreferenceIcon(bool? preference)
+    {
+        return preference == true ? "bi-check-lg" : "bi-x-lg";
+    }
+}
+
+@if (Model.TenantList!.Count == 0)
```
oskarwilliams378, line 16: `!` not needed :)

web/Views/Admin/_MatchListTable.cshtml
```diff
@@ -0,0 +1,82 @@
...
+{
+    <h2 class="font-h3">There are no tenants which match these search criteria.</h2>
+    return;
```
oskarwilliams378, line 19: No return needed either :)

web/Views/Admin/_MatchListTable.cshtml
```diff
@@ -0,0 +1,82 @@
...
+            <th scope="col">Email</th>
+            <th scope="col">Phone</th>
+            <th scope="col">Postcode</th>
+            <th scope="col">Type</th>
+            <th scope="col">Criteria</th>
+            <th scope="col">Match</th>
+        </tr>
+        </thead>
+        <tbody>
+        @foreach (var tenant in Model.TenantList!)
```
oskarwilliams378, line 36: Can remove the ! here :)

web/Views/Admin/_MatchListTable.cshtml
```diff
@@ -0,0 +1,82 @@
...
+                        On housing benefits
+                    </div>
+                    <div class="@GetPreferenceFont(tenant.ETT)">
+                        <i class="bi me-1 @GetPreferenceIcon(tenant.ETT)"></i>
+                        In EET (Employment, Education, or Training)
+                    </div>
+                </td>
+                <td>
+                    @using (Html.BeginForm("SendMatchLinkEmail", "Admin", new
+                        {
```
oskarwilliams378, line 68: Weird indentation here :)

web/Views/Property/ViewProperty.cshtml
```diff
@@ -26,6 +26,13 @@
     </div>
 
     <div class="col-md d-flex flex-column pb-2">
+        <div class="d-flex">
+            <h3 class="font-h3 d-inline-block">@(User.IsInRole("Landlord") || User.IsInRole("Admin") ? "Tenant preferences" : "Suitable for")</h3>
+            @if (User.IsInRole("Landlord") || User.IsInRole("Admin"))
```
oskarwilliams378, line 31: I've seen this in a lot of different places, and I think this is an example where adding an extension to the ClaimsPrincipal class (what User is) that is something like `IsLandlordOrAdmin` :)

oskarwilliams378 (changes requested):

web/Views/Admin/_MatchListTable.cshtml
TomHanmer29: I think we want the return here so that if there are no tenants it doesn't display the table headers.

oskarwilliams378: Sure

web/CustomExtensions/UserExtensions.cs
```diff
@@ -0,0 +1,12 @@
...
+namespace BricksAndHearts.CustomExtensions
+{
+    public static class UserExtensions
+    {
+        public static bool IsLandlordOrAdmin(this ClaimsPrincipal user)
+        {
+            return user.IsInRole("Landlord") || user.IsInRole("Admin");
+        }
+    }   
+}
```
oskarwilliams378, line 12: We can use file scoping as we do elsewhere :) 
```suggestionusing System.Security.Claims;
namespace BricksAndHearts.CustomExtensions;

public static class UserExtensions
{
    public static bool IsLandlordOrAdmin(this ClaimsPrincipal user)
    {
        return user.IsInRole("Landlord") || user.IsInRole("Admin");
    }
}

```

oskarwilliams378 (approved):

SleepyUser (approved):

PR #172, opened 2022-08-23
BNH-180: Fixes surrounding availability datepicker

SleepyUser:
-Datepicker now only shows date and not time
-Datepicker now correctly shows when AvailableSoon is selected
-User is only required to enter a date on step six if AvailableSoon is selected
-User may only choose a future date client side.
-Server side, both validation rules above are enforced
-Fixed an issue with server side invalid model redirects for steps 3-5 always redirecting to step 2

### DIFF
--- a/web/wwwroot/js/site.js
+++ b/web/wwwroot/js/site.js
@@ -17,7 +17,6 @@ function checkIfSelectedValueIsTarget(target, id) {
     }
 }
 
-
 function makeRequiredIfSelectedValueIsTarget(target, id, idToRequire) {
     if ($("#" + id + " :selected").val() == target) {
         $("#" + idToRequire).prop('required', true);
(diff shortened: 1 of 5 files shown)


emilypatsko:
I think this looks good, could you add some screenshots/similar to show how it actually looks when 'available soon' is/isn't selected?

emilypatsko (commented):
I think this looks good, could you add some screenshots/similar to show how it actually looks when 'available soon' is/isn't selected?

web/wwwroot/js/site.js
```diff
@@ -17,8 +17,17 @@ function checkIfSelectedValueIsTarget(target, id){
     }
 }
 
+
```
oskarwilliams378, line 20: extra new line

oskarwilliams378 (commented):
One minor comment

SleepyUser:
Any other option selected:
image
Available soon selected:
image
(Because I'm happy it works, bonus screenshot of previous dates being greyed out)
image

oskarwilliams378 (approved):

PR #174, opened 2022-08-23
BNH-191 - landlord registration changes

cec200wire:
A few minor changes:
 - asterisks added to show which fields are compulsory (and optional labels therefore removed)
 - text added to Address page to instruct landlords to provide personal address (rather than business or property)
Both those changes visible on this page:
image

 - the 'title' dropdown now has an 'Other' option, which brings up a text box in which the user enters their desired title. This is a compulsory field if and only if 'Other' has been selected.
image

### DIFF
--- a/web/ViewModels/LandlordProfileModel.cs
+++ b/web/ViewModels/LandlordProfileModel.cs
@@ -14,6 +14,8 @@ public class LandlordProfileModel
     [DisplayName("Title")]
     public string Title { get; set; } = string.Empty;
 
+    public string? TitleInput { get; set; }
+
     [Required]
     [StringLength(255)]
     [DisplayName("First name")]
--- a/web/Views/Landlord/_RegisterPagePartial.cshtml
+++ b/web/Views/Landlord/_RegisterPagePartial.cshtml
@@ -16,63 +16,71 @@
 
     <div class="my-3">
         @Html.HiddenFor(m => m.Unassigned, new { value = Model.Unassigned })
-        <label asp-for="Title" class="form-label"></label>
-        <select asp-for="Title" class="form-select" required>
+        <label asp-for="Title" class="form-label">Title*</label>
+        <select asp-for="Title" class="form-select" required
+                id="Title" onchange="displayIfSelectedValueIsTarget('Other', 'Title', 'titleOther')">
             <option selected disabled value="">Title</option>
             <option value="Mr">Mr</option>
             <option value="Mrs">Mrs</option>
             <option value="Miss">Miss</option>
             <option value="Ms">Ms</option>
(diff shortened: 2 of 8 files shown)


web/wwwroot/js/site.js
```diff
@@ -25,14 +25,6 @@ function makeRequiredIfSelectedValueIsTarget(target, id, idToRequire) {
     }
 }
 
-function insertDefaultPostcodeIfNotSortByLocation(id) {
```
oskarwilliams378, line 28: was this meant to be removed?

oskarwilliams378 (changes requested):

web/wwwroot/js/site.js
cec200wire: It didn't seem to be being used, so I removed it - but I've just check and it is being used (and I've also realised I slightly broke something else, so I'm going to fix that quickly!)

oskarwilliams378 (approved):

web/ViewModels/LandlordProfileModel.cs
```diff
@@ -14,6 +14,8 @@ public class LandlordProfileModel
     [DisplayName("Title")]
     public string Title { get; set; } = string.Empty;
 
+    public string? TitleInput { get; set; }
```
MaxTheComputerer, line 17: The Title and Other fields won't be filled in correctly when you edit a landlord because they're not mapped directly to the database model anymore.

One way around this would be to keep the server-side code as just having `Title` and not `TitleInput` and then use some Javascript stuff when the form is submitted to put the correct value into `Title`.

Or we could do it this way but then add some extra logic into the the `fromDbModel` to fill `Title` and `TitleInput`.

@oskarwilliams378 what do you think?

web/Views/Landlord/_RegisterPagePartial.cshtml
```diff
@@ -16,63 +16,71 @@
...
             <option selected disabled value="">Title</option>
             <option value="Mr">Mr</option>
             <option value="Mrs">Mrs</option>
             <option value="Miss">Miss</option>
             <option value="Ms">Ms</option>
             <option value="Dr">Dr</option>
+            <option value="Prof">Prof</option>
+            <option value="Other">Other</option>
         </select>
+        <div id="titleOther" class="my-3" style="@(Model.MembershipId == null ? "display: none;" : "")">
```
MaxTheComputerer, line 31: This should probably be `Model.TitleInput` instead of `Model.MembershipId`? Because at the moment, when you edit a landlord, the "other" input is always shown, even when the title is "Mr" etc.

MaxTheComputerer (changes requested):

PR #178, opened 2022-08-23
BNH-193 Make routes more clearly match site navigation

MaxTheComputerer:
- Routes now follow the site navigation
- All admin pages start with `/admin`, all landlord pages start with `/landlord`, and property pages start with `/property`
- The lists pages are of the form `/admin/lists/XX`
- Operations on these pages follow this pattern, e.g. `/admin/lists/tenants/import`
- Individual landlords' dashboards are no longer accessible by admins

### DIFF
--- a/web/Controllers/HomeController.cs
+++ b/web/Controllers/HomeController.cs
@@ -22,12 +22,12 @@ public class HomeController : AbstractController
             var user = CurrentUser;
             if (user.IsAdmin)
             {
-                return RedirectToAction("AdminDashboard", "Admin");
+                return RedirectToAction(nameof(AdminController.AdminDashboard), "Admin");
             }
 
             if (user.LandlordId != null)
             {
-                return RedirectToAction("Dashboard", "Landlord");
+                return RedirectToAction(nameof(LandlordController.Dashboard), "Landlord");
             }
         }
 
--- a/web/Controllers/HomeController.cs
+++ b/web/Controllers/HomeController.cs
@@ -27,7 +27,7 @@ public class HomeController : AbstractController
 
             if (user.LandlordId != null)
             {
-                return RedirectToAction("MyDashboard", "Landlord");
+                return RedirectToAction("Dashboard", "Landlord");
             }
         }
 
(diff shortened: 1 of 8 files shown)


web/Controllers/HomeController.cs
```diff
@@ -27,7 +27,7 @@ public IActionResult Index()
 
             if (user.LandlordId != null)
             {
-                return RedirectToAction("MyDashboard", "Landlord");
+                return RedirectToAction("Dashboard", "Landlord");
```
oskarwilliams378, line 30: One thing to try and would make sure all of these are for correct methods is name the actions as nameof(LandlordController.Dashboard) etc :)

web/Controllers/LandlordController.cs
```diff
@@ -319,16 +324,16 @@ public async Task<IActionResult> TieUserWithLandlord(string inviteLink)
...
     }
 
-    [HttpGet("{id:int}/dashboard")]
-    public async Task<ActionResult> Dashboard([FromRoute] int id)
+    [HttpGet("dashboard")]
+    public async Task<ActionResult> Dashboard()
     {
         var user = CurrentUser;
-        if (user.LandlordId != id && !user.IsAdmin)
+        if (!user.LandlordId.HasValue)
```
oskarwilliams378, line 331: what about for admins?

oskarwilliams378 (changes requested):

web/Controllers/LandlordController.cs
MaxTheComputerer: This is for the landlord dashboard which I figured admins shouldn't ever need to access (because they have their own dashboard and other pages for dealing with landlord profiles). For some reason the existing code allowed admins to see the dashboard of any landlord, but I'm not sure that makes sense to do, and we never made use of it anyway.

web/Controllers/HomeController.cs
MaxTheComputerer: Oh yeah that's a good idea!

web/Controllers/LandlordController.cs
oskarwilliams378: Sounds reasoning :)

oskarwilliams378 (approved):

PR #179, opened 2022-08-23
BNH-176: Ensuring server-side validation and buttons on tenant preference form page function correctly

SleepyUser:
-Users can no longer submit their preferences without selecting at least one of Single, Couple, Family
image

-Users no longer have their form page reset to blank if the model is invalid 
-The Single, Couple, Family buttons now style correctly, appearing filled in with colour if the underlying value is true.
image

### DIFF
--- a/web/Views/Property/PropertyInputForm/TenantPreferences.cshtml
+++ b/web/Views/Property/PropertyInputForm/TenantPreferences.cshtml
@@ -7,15 +7,15 @@
 
     <div class="row g-3">
         <div class="col-auto d-grid">
-            @Html.CheckBox("AcceptsSingleTenant", Model.AcceptsSingleTenant ?? false, new { @class = "btn-check" })
+            <input type="checkbox" asp-for="AcceptsSingleTenant" value="@(Model.AcceptsSingleTenant.ToString())" class="check-min-one" onchange="ifNoneCheckedAddRequiredIfAnyCheckedRemoveRequired()"/>
             <label asp-for="AcceptsSingleTenant" class="btn btn-outline-blue shadow-none pill-check">Single tenant</label>
         </div>
         <div class="col-auto d-grid">
-            @Html.CheckBox("AcceptsCouple", Model.AcceptsCouple ?? false, new { @class = "btn-check" })
+            <input type="checkbox" asp-for="AcceptsCouple" value="@(Model.AcceptsCouple.ToString())" class="check-min-one" onchange="ifNoneCheckedAddRequiredIfAnyCheckedRemoveRequired()"/>
             <label asp-for="AcceptsCouple" class="btn btn-outline-blue shadow-none pill-check">Couple</label>
         </div>
         <div class="col-auto d-grid">
-            @Html.CheckBox("AcceptsFamily", Model.AcceptsFamily ?? false, new { @class = "btn-check" })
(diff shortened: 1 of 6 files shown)


web/Views/Property/PropertyInputForm/TenantPreferences.cshtml
```diff
@@ -7,15 +7,15 @@
...
         <div class="col-auto d-grid">
-            @Html.CheckBox("AcceptsSingleTenant", Model.AcceptsSingleTenant ?? false, new { @class = "btn-check" })
-            <label asp-for="AcceptsSingleTenant" class="btn btn-outline-blue shadow-none pill-check">Single tenant</label>
+            <input type="checkbox" asp-for="AcceptsSingleTenant" class="btn-check"/>
+            <label asp-for="AcceptsSingleTenant" id="AcceptsSingleTenant" class="btn btn-outline-blue shadow-none pill-check">Single tenant</label>
         </div>
         <div class="col-auto d-grid">
-            @Html.CheckBox("AcceptsCouple", Model.AcceptsCouple ?? false, new { @class = "btn-check" })
+            <input type="checkbox" asp-for="AcceptsCouple" class="btn-check"/>
             <label asp-for="AcceptsCouple" class="btn btn-outline-blue shadow-none pill-check">Couple</label>
```
oskarwilliams378, line 15: Should these labels gain IDs too?

web/wwwroot/js/site.js
```diff
@@ -25,6 +25,7 @@ function makeRequiredIfSelectedValueIsTarget(target, id, idToRequire) {
     }
 }
 
+
```
oskarwilliams378, line 28: extra new line

web/Views/Property/PropertyInputForm/TenantPreferences.cshtml
```diff
@@ -7,15 +7,15 @@
 
     <div class="row g-3">
         <div class="col-auto d-grid">
-            @Html.CheckBox("AcceptsSingleTenant", Model.AcceptsSingleTenant ?? false, new { @class = "btn-check" })
-            <label asp-for="AcceptsSingleTenant" class="btn btn-outline-blue shadow-none pill-check">Single tenant</label>
+            <input type="checkbox" asp-for="AcceptsSingleTenant" class="btn-check"/>
```
oskarwilliams378, line 10: Happy to use tags, we can also use Html.CheckBoxFor if you prefer the other method (I am very happy either way :) )

oskarwilliams378 (changes requested):

web/Views/Property/PropertyInputForm/TenantPreferences.cshtml
SleepyUser: The labels are the buttons, so the asp-for is intended to give them the id and name to allow the user to click it and change the underlying boolean. The input checkbox is hidden, but linked to the same thing, allowing the CSS to look for "input:checked ~ label" and show the label button as pressed.

oskarwilliams378: In which case should we remove the ID from the button above?

SleepyUser: Yes! :) fixed

oskarwilliams378 (approved):

PR #180, opened 2022-08-23
BNH-191 - Landlord Title fix

cec200wire:
The landlord Title dropdown, and Title (other) box, now populate with the correct information when the Title is non-standard. Also, this non-standard title will display on the Review page (as opposed to simply other).

The Title (other) box only appears on first loading the edit form if the existing title is in the category Other.

### DIFF
--- a/web/ViewModels/LandlordProfileModel.cs
+++ b/web/ViewModels/LandlordProfileModel.cs
@@ -14,6 +14,8 @@ public class LandlordProfileModel
     [DisplayName("Title")]
     public string Title { get; set; } = string.Empty;
 
+    public static readonly string[] KnownTitles = { "Mr", "Mrs", "Miss", "Ms", "Dr", "Prof" };
+
     public string? TitleInput { get; set; }
 
     [Required]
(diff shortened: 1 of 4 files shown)


MaxTheComputerer (dismissed):
Looks good!

web/ViewModels/LandlordProfileModel.cs
```diff
@@ -86,5 +85,22 @@ public static LandlordProfileModel FromDbModel(LandlordDbModel landlord)
                 Postcode = landlord.Postcode
             }
         };
+
+        if (landlord.Title != "Mr"
+            && landlord.Title != "Mrs"
+            && landlord.Title != "Miss"
+            && landlord.Title != "Ms"
+            && landlord.Title != "Dr"
+            && landlord.Title != "Prof")
```
oskarwilliams378, line 94: We can simplify this slightly :) 

```suggestion
        if (new[] { "Mr", "Mrs", "Miss", "Ms", "Dr", "Prof" }.Contains(landlord.Title))

```

web/ViewModels/LandlordProfileModel.cs
```diff
@@ -60,10 +60,9 @@ public class LandlordProfileModel
```
oskarwilliams378, line 60: So I think it may be worth adding the list of possible titles to this model, instead of having it as a magic string. We can then loop over that list in the view for the options, and have it here for the check on line 89 as well

oskarwilliams378 (changes requested):

oskarwilliams378 (approved):

PR #183, opened 2022-08-24
Bnh 194 navbar restyling for mobiles

TomHanmer29:
navbar looks the same on desktop
image

unexpanded on mobile

image

expanded on mobile

image

(double navbars in this image, imagine only 1 blue navbar is present if a user is not both a landlord and admin)

### DIFF
--- a/web/Views/Shared/_Layout.cshtml
+++ b/web/Views/Shared/_Layout.cshtml
@@ -34,18 +34,20 @@
                 <div class="navbar-nav">
                     @if (User.Identity?.IsAuthenticated == true)
                     {
-                        @if (User.IsInRole("Landlord"))
-                        {
-                            <div class="d-flex align-items-center mx-3 me-sm-0 py-2 py-sm-0">
-                                <a asp-controller="Landlord" asp-action="Profile" asp-route-id="@(((BricksAndHeartsUser)User.Identity).LandlordId!.Value)" class="profile-photo-sm">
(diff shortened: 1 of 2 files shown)


web/Views/Shared/_Layout.cshtml
```diff
@@ -34,18 +34,20 @@
...
-                                    <img src="@(((BricksAndHeartsUser)User.Identity).GoogleProfileImageUrl ?? Url.Content("~/images/profileDefault.png"))" alt="Profile picture" referrerpolicy="no-referrer"/>
-                                </a>
-                            </div>
-                        }
-                        @using (Html.BeginForm("Logout", "Login", FormMethod.Post))
-                        {
-                            <button type="submit" class="btn btn-thin btn-outline-primary mx-3">Log out</button>
-                        }
+                        <div class="d-flex">
+                            @if (User.IsInRole("Landlord"))
```
oskarwilliams378, line 38: indentation is off here I thinkk

oskarwilliams378 (changes requested):

web/Views/Shared/_Layout.cshtml
TomHanmer29: I thought if we have a div in an if statement we were supposed to indent it? What would you change it to be?

TomHanmer29: and similarly for if statements in div's

TomHanmer29: ahhh ok thanks for clearing that up :)

oskarwilliams378 (approved):

PR #184, opened 2022-08-24
BNH-189 Property table restyling

MaxTheComputerer:
Updated property table to reflect feedback from UAT.

- Refactored `PropertyListModel` to have a list of `PropertyDetailsModel`s so we can access images and landlord data
- Added images to property table
- Changed description to hide overflow and link to details page to read all of it
- Changed "View" buttons to "View/Edit"

image

### DIFF
--- a/web/Views/Admin/PropertyList.cshtml
+++ b/web/Views/Admin/PropertyList.cshtml
@@ -20,7 +20,7 @@
                     },
                     new { @class = "form-select col h-auto me-3", onchange = "displayIfSelectedValueIsTarget('Location', 'SortBy', 'postcode'); insertDefaultPostcodeIfNotSortByLocation(id)" })
                 <div id="postcode" class="col-auto" style="@(Model.SortBy != "Location" ? "display: none;" : "")">
-                    <input class="form-control col h-auto me-3" id="target" name="target" required placeholder="Enter postcode to sort by location" pattern="@Constants.PostcodeValidationRegex"/>
+                    <input class="form-control col h-auto me-3" id="target" name="target" required placeholder="Enter postcode to sort by location" pattern="@Constants.PostcodeValidationRegex" value="@(Model.SortBy == "Location" ? "" : "CB2 1LA")"/>
                 </div>
                 <button type="submit" class="btn btn-outline-primary col">Sort</button>
             </form>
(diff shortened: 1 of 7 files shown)


web/Views/Admin/PropertyList.cshtml
```diff
@@ -20,7 +20,7 @@
                     },
                     new { @class = "form-select col h-auto me-3", onchange = "displayIfSelectedValueIsTarget('Location', 'SortBy', 'postcode'); insertDefaultPostcodeIfNotSortByLocation(id)" })
                 <div id="postcode" class="col-auto" style="@(Model.SortBy != "Location" ? "display: none;" : "")">
-                    <input class="form-control col h-auto me-3" id="target" name="target" required placeholder="Enter postcode to sort by location" pattern="@Constants.PostcodeValidationRegex"/>
+                    <input class="form-control col h-auto me-3" id="target" name="target" required placeholder="Enter postcode to sort by location" pattern="@Constants.PostcodeValidationRegex" value="@(Model.SortBy == "Location" ? "" : "CB2 1LA")"/>
```
cec200wire, line 23: Not anything to do with your code really, but we should probably change this placeholder postcode to be ChangeAhead's postcode rather than Softwire's

cec200wire (commented):
All seems good - passing landlord details into the property table will be very useful for redefining the Status in a more useful manner.

web/Views/Property/_PropertyTable.cshtml
```diff
@@ -28,12 +28,23 @@
...
         {
+            var prop = propertyDetails.Property!;
             <tr>
                 <td class="text-nowrap">
                     <partial name="../Property/Details/_AvailabilityPill" model="prop"/>
                 </td>
+                <td>
+                    @if (propertyDetails.Images.Count > 0)
+                    {
+                        var image = propertyDetails.Images.First();
```
oskarwilliams378, line 41: You could do these together with 
```suggestion
                    @if (propertyDetails.Images.FirstOrDefault() is { } image)
                    {

```

oskarwilliams378 (changes requested):

web/Views/Admin/PropertyList.cshtml
MaxTheComputerer: That's a good point, I've changed it now!

oskarwilliams378 (approved):

PR #189, opened 2022-08-24
BNH-122 - refine availability state/'live' logic

cec200wire:
[This may cause some problems initially for already-existing databases, until 'Available soon' is changed to 'Available']

**Availability states:**
•	‘Draft’ status, on the property form, now triggers a message explaining what this status is (it confused at least one user in UAT testing)
•	‘AvailableSoon’ status no longer exists in the database. Instead, ‘Available’ with an ‘AvailableFrom’ date in the future corresponds to ‘AvailableSoon’, while with a date in the past or no date corresponds to ‘Available’. The two states are distinguished in View Models and in the Property Form, for the convenience of users; but by combining them in the database, ‘AvailableSoon’ basically converts automatically to ‘Available’ once the relevant date is reached.
•	‘Occupied’ and ‘RenterUserId’ both appear to not be being accessed in any way at the moment, so there is nowhere to implement the check that if ‘Occupied’ is true, ‘RenterUserId’ should be non-null

**‘Live’ status and Property Lists:**
A property is ‘Live’ if:
- ­	‘AvailabilityState’ is not ‘Draft’
- ­	Landlord’s ‘CharterApproved’ is ‘true’
- ­	Landlord’s ‘Disabled’ is ‘false’
All other properties will be marked as ‘Pending’, where such a label is needed

On the landlord-specific Property Lists, all properties owned by that landlord appear
- ­	If an Admin is viewing the properties of a specific landlord, it is likely that they are also aware of their ‘CharterApproved’ and ‘Disabled’ states, so these need not potentially empty the entire list
- ­	‘Draft’ properties appear on this list for the convenience of a Landlord; admins can see them in this way if they make an effort to 

On the general Property List (visible only to admins)

- ­	‘Draft’ properties do not appear: this Availability State exists in order to facilitate Landlords having unfinished listings, or listings which for other reasons they do not want to be live; therefore, they should not appear on a list of properties which admins may be taking actions (such as finding potential matches) on
- ­	Properties owned by disabled Landlords do not appear: if a landlord has been set as ‘Disabled’, there is likely to be a good reason for this, and so their properties should be removed from the circulation of matching and occupation
- ­	Properties owned by unapproved Landlords do appear: these Landlords are just awaiting approval, rather than having any negative condition imposed on them, so their properties can begin to be considered; in practical terms, seeing that a Landlord who is awaiting Charter Approval has desirable properties may encourage some effort to be made to advance the approval process.


web/ViewModels/PropertyInput/PropertyInputModelAvailability.cs
```diff
@@ -1,71 +1,75 @@
...
-        return new PropertyViewModel
-        {
-            Availability = Availability,
-            Rent = Rent,
-            AvailableFrom = AvailableFrom,
-            TotalUnits = TotalUnits,
-            OccupiedUnits = OccupiedUnits
-        };
-    }
+using System.ComponentModel.DataAnnotations;
```
oskarwilliams378, line 1: Again something weird is happening here, why has the whole file come into git?

web/ViewModels/PropertyViewModel.cs
```diff
@@ -114,11 +114,18 @@ public static PropertyViewModel FromDbModel(PropertyDbModel property)
...
                 AcceptsWithoutGuarantor = property.AcceptsWithoutGuarantor
             },
-            Availability = property.Availability,
-            AvailableFrom = property.AvailableFrom,
             TotalUnits = property.TotalUnits,
             OccupiedUnits = property.OccupiedUnits,
             PublicViewLink = property.PublicViewLink
         };
+
+        propertyViewModel.Availability =
```
oskarwilliams378, line 122: Similar to the comment in the service, could we make this a variable set before the model and then we can set these in the object initialisation

web/Views/Property/Details/_LivePill.cshtml
```diff
@@ -0,0 +1,21 @@
...
+        <span class="status-pill-@colourClass rounded-pill px-3 py-1">
+            <i class="bi bi-@icon me-1"></i>
+            @availability
+        </span>
+    }
+}
+
+@if (Model.Property != null && Model.Property.Availability != AvailabilityState.Draft && Model.Owner != null && Model.Owner.CharterApproved && !Model.Owner.Disabled)
+{
+    GetLivePill("Live", "live","eye");
```
oskarwilliams378, line 16:
```suggestion
    GetLivePill("Live", "live", "eye");

```

web/Views/Property/Details/_LivePill.cshtml
```diff
@@ -0,0 +1,21 @@
...
+}
+else
+{
+    GetLivePill("Pending", "draft","clock");
```
oskarwilliams378, line 20:
```suggestion
    GetLivePill("Pending", "draft", "clock");

```

web/Views/Property/_PropertyTable.cshtml
```diff
@@ -62,11 +62,11 @@
                 <td>
                     @if (prop.Availability == AvailabilityState.AvailableSoon && prop.AvailableFrom.HasValue)
                     {
-                        @("From " + prop.AvailableFrom.Value.ToShortDateString())
+                        @("Available" + "\n" + prop.AvailableFrom.Value.ToShortDateString())
```
oskarwilliams378, line 65: Should this say Available from instead of Available?

oskarwilliams378 (changes requested):

web/Services/PropertyService.cs
```diff
@@ -68,15 +68,20 @@ public async Task<int> AddNewProperty(int landlordId, PropertyViewModel createMo
...
             Rent = createModel.Rent,
-
-            Availability = createModel.Availability ?? AvailabilityState.Draft,
+            
             TotalUnits = createModel.TotalUnits ?? 1,
             OccupiedUnits = createModel.OccupiedUnits ?? 0,
-            AvailableFrom = createModel.Availability == AvailabilityState.AvailableSoon
-                ? createModel.AvailableFrom
-                : null
+
```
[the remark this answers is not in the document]
cec200wire: Line numbering seems to have gone slightly off for me, but I think I've removed the blank line.

web/Views/Property/_PropertyTable.cshtml
cec200wire: I don't think it makes a massive difference - 'Available' followed by a date I think fairly clearly implies the 'from' part, but I can add it.

web/Services/PropertyService.cs
```diff
@@ -127,21 +132,13 @@ public async Task UpdateProperty(int propertyId, PropertyViewModel updateModel,
...
-            // If we're not "available soon" then don't have a from date
-            dbModel.AvailableFrom = null;
-        }
+            : updateModel.Availability == AvailabilityState.AvailableSoon
+                ? AvailabilityState.Available
+                : updateModel.Availability ?? AvailabilityState.Draft;
+        
+        dbModel.AvailableFrom = dbModel.Availability == AvailabilityState.Available
+            ? updateModel.Availability == AvailabilityState.AvailableSoon? updateModel.AvailableFrom : DateTime.Now
+            : null;
```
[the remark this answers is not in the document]
cec200wire: Yes, I did think it was a bit confusing, although I wasn't sure if that was just my unfamiliarity with ternaries making me take longer to process it -I'll split it into two parts.

web/Services/PropertyService.cs
```diff
@@ -125,24 +131,21 @@ public async Task UpdateProperty(int propertyId, PropertyViewModel updateModel,
...
-            // If we're not "available soon" then don't have a from date
-            dbModel.AvailableFrom = null;
+            dbModel.Availability = updateModel.Availability == AvailabilityState.AvailableSoon
+                ? AvailabilityState.Available
+                : updateModel.Availability ?? AvailabilityState.Draft;
         }
 
+        dbModel.AvailableFrom = dbModel.Availability == AvailabilityState.Available
+            ? updateModel.Availability == AvailabilityState.AvailableSoon? updateModel.AvailableFrom : DateTime.Now
+            : null;
```
oskarwilliams378, line 147: still a lot of nested ternaries here, which may be worth expanding

web/Services/PropertyService.cs
```diff
@@ -68,13 +72,15 @@ public async Task<int> AddNewProperty(int landlordId, PropertyViewModel createMo
...
+            
             TotalUnits = createModel.TotalUnits ?? 1,
             OccupiedUnits = createModel.OccupiedUnits ?? 0,
-            AvailableFrom = createModel.Availability == AvailabilityState.AvailableSoon
-                ? createModel.AvailableFrom
-                : null
+            
+            Availability = createdAvailability,
+            AvailableFrom = createdAvailability == AvailabilityState.Available
+            ? createModel.Availability == AvailabilityState.AvailableSoon
```
oskarwilliams378, line 81: indenatation has gone weird here

oskarwilliams378 (changes requested):

web/Services/PropertyService.cs
cec200wire: The indentation is probably weird because of the nested ternary - I'm not even sure how this should be indented to begin with, I though it looked right the way it currently is.

web/Services/PropertyService.cs
cec200wire: This nested ternary is directly equivalent to the one in the object initialiser in AddNewProperty, so I don't think it makes sense to expand this one without also expanding the other one (it would disguise their similarities); if I expand both of them, then I believe the one in AddNewProperty would have to move out of the object initialiser. I'm not sure that the benefits of expanding the ternary are significant enough to justify that disadvantage, although if you strongly think that it should be expanded, it can be.

cec200wire:
I've moved the conversion of Available Soon (in View) to Available (for DB) into the View Model, along with the Occupied Units logic. This broke a couple of unit tests, which were checking the Occupied units logic - there wasn't an immediately obvious way to fix them, so I've just commented them out for now as they are testing functionality that hasn't actually been implemented yet (there is no way to input the data which would trigger these cases, so they don't need checking).

web/ViewModels/PropertyInput/PropertyInputModelAvailability.cs
```diff
@@ -1,71 +1,97 @@
...
+        {
+            dbAvailability = AvailabilityState.Occupied;
+        }
+        else
+        {
+            dbAvailability = Availability == AvailabilityState.AvailableSoon
+                ? AvailabilityState.Available : Availability;
+        }
+
+        var dbAvailableFrom = new DateTime?();
```
oskarwilliams378, line 77:
```suggestion
        DateTime? dbAvailableFrom = null;

```

If we doe this we can remove the else below

BricksAndHearts.UnitTests/ServiceTests/Property/PropertyServiceTests.cs
```diff
@@ -147,7 +147,8 @@ public async void UpdateProperty_UpdatesProperty()
         property.AcceptsBenefits.Should().BeFalse();
         context.Properties.Count().Should().Be(propertiesBeforeCount);
     }
-
+    
+    /* This logic now occurs in the View Model, causing the test to fail
```
oskarwilliams378, line 151: Instead of commenting we should probably remove

cec200wire: I've rewritten it to test the logic in the View Model (I thought it would be more complicated than it was)

oskarwilliams378 (approved):

PR #192, opened 2022-08-25
BNH-193 Routing bug fix

MaxTheComputerer:
The "view properties" button on the landlord profile will take you to `/landlord/{id}/properties` if you're an admin, and `/landlord/me/properties` if you're a landlord. The admin behaviour overrides the landlord behaviour if you're both.

### DIFF
--- a/web/Views/Landlord/Profile.cshtml
+++ b/web/Views/Landlord/Profile.cshtml
@@ -133,7 +133,7 @@
 
 @if (Model.NumOfProperties > 0)
 {
-    <a asp-action="@(User.IsInRole("Landlord") ? "ViewMyProperties" : "ViewProperties")" asp-controller="Landlord" class="btn btn-primary my-3">View properties</a>
+    <a asp-action="@(User.IsInRole("Admin") ? "ViewProperties" : "ViewMyProperties")" asp-controller="Landlord" asp-route-id="@(User.IsInRole("Admin") ? Model.LandlordId : "")" class="btn btn-primary my-3">View properties</a>
 }
 else
 {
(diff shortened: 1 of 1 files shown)


web/Views/Landlord/Profile.cshtml
```diff
@@ -133,7 +133,7 @@
 
 @if (Model.NumOfProperties > 0)
 {
-    <a asp-action="@(User.IsInRole("Landlord") ? "ViewMyProperties" : "ViewProperties")" asp-controller="Landlord" class="btn btn-primary my-3">View properties</a>
+    <a asp-action="@(User.IsInRole("Admin") ? "ViewProperties" : "ViewMyProperties")" asp-controller="Landlord" asp-route-id="@(User.IsInRole("Admin") ? Model.LandlordId : "")" class="btn btn-primary my-3">View properties</a>
```
oskarwilliams378, line 136: As an Admin would this not take me to the View Properties for myself if I was also a Landlord?

oskarwilliams378 (changes requested):

web/Views/Landlord/Profile.cshtml
MaxTheComputerer: I don't think so - if you're an admin, then it goes to `ViewProperties`, if you're not an admin then it goes to `ViewMyProperties`. So if you're both then it should do the latter.

oskarwilliams378: Not sure what you mean. If you are an admin, and on your own page, it will still go to `ViewProperties`

oskarwilliams378 (changes requested):

oskarwilliams378 (approved):

PR #197, opened 2022-08-25
BNH-205 added without guarantor to matching and filtering

TomHanmer29:
added an extra database field, filtering button, and extra filters on the matching data.
image
image


BricksAndHearts.UnitTests/ServiceTests/CsvImport/CsvImportServiceTests.cs
```diff
@@ -34,7 +34,7 @@ public async void CheckIfImportWorks_CalledOnFileWithMissingColumn_ReturnsDanger
         await using var context = Fixture.CreateReadContext();
         var service = new CsvImportService(context, PostcodeService, Logger);
 
-        var data = Encoding.UTF8.GetBytes("Name,Phone,Postcode,Type,HasPet,NotInEET,UniversalCredit,HousingBenefits,Under35");
+        var data = Encoding.UTF8.GetBytes("Name,Phone,Postcode,Type,HasPet,NotInEET,UniversalCredit,HousingBenefits,Under35.NoGuarantor");
```
emilypatsko, line 37: Should that be a comma instead of a full stop?

oskarwilliams378, line 37: should this be 
```suggestion
        var data = Encoding.UTF8.GetBytes("Name,Phone,Postcode,Type,HasPet,NotInEET,UniversalCredit,HousingBenefits,Under35,NoGuarantor");

```

oskarwilliams378 (changes requested):

web/Views/Admin/_MatchListTable.cshtml
```diff
@@ -62,6 +62,10 @@
                         <i class="bi me-1 @GetPreferenceIcon(!tenant.NotInEET)"></i>
                         In EET (Employment, Education, or Training)
                     </div>
+                    <div class="@GetPreferenceFont(tenant.NoGuarantor)">
+                        <i class="bi me-1 @GetPreferenceIcon(tenant.NoGuarantor)"></i>
+                        Without rent guarantor
```
emilypatsko, line 67: Can we make this 'Has rent guarantor' instead? I think that's the default assumption, and the double negative (⛔ Without rent guarantor) is confusing

web/Views/Admin/_TenantTable.cshtml
```diff
@@ -61,6 +61,10 @@
                         <i class="bi me-1 @GetPreferenceIcon(!tenant.NotInEET)"></i>
                         In EET (Employment, Education, or Training)
                     </div>
+                    <div class="@GetPreferenceFont(tenant.NoGuarantor)">
+                        <i class="bi me-1 @GetPreferenceIcon(tenant.NoGuarantor)"></i>
+                        Without rent guarantor
```
emilypatsko, line 66: Same here

web/Database/TenantDbModel.cs
```diff
@@ -19,4 +19,5 @@ public class TenantDbModel
     public bool? UniversalCredit { get; set; }
     public bool? HousingBenefits { get; set; }
     public bool? Under35 { get; set; }
+    public bool? NoGuarantor { get; set; }
```
emilypatsko, line 22: I think this should be 'HasGuarantor' - basically flipping the bool on its head. I know we'd like to make it so all the bools default to false, but I think that's just confusing for some of them (especially when it results in a double negative)

web/Views/Property/PropertyInputForm/TenantPreferences.cshtml
```diff
@@ -45,7 +45,7 @@
     <div class="form-check">
         <input type="checkbox" asp-for="AcceptsBenefits" class="form-check-input"/>
         <label asp-for="AcceptsBenefits" class="form-check-label">
-            on Universal Credit or housing benefits?
+            on Housing Benefits?
```
emilypatsko, line 48: What's the reason for this change?

web/Views/Shared/Details/_TenantPreferences.cshtml
```diff
@@ -35,7 +35,7 @@
         {
             <div class="@GetPreferenceFont(Model.Property.LandlordRequirements.AcceptsUnder35)">
                 <i class="bi me-1 @GetPreferenceIcon(Model.Property.LandlordRequirements.AcceptsUnder35)"></i>
-                Is over 35
+                Is under 35
```
emilypatsko, line 38: What's the reason for this change? Is this one that was missed out before?

emilypatsko (changes requested):

BricksAndHearts.UnitTests/ServiceTests/CsvImport/CsvImportServiceTests.cs
emilypatsko: I think @oskarwilliams378 beat me to it...

emilypatsko:
Also, could we make it so that when viewing a property (including on the matching page), the requisite _for the property_ is shown as 

👍🏻 Accepts without guarantor
👎🏻 Accepts without guarantor

(obviously not with thumbs up/down but I couldn't find tick/cross emojis)

web/Views/Shared/Details/_TenantPreferences.cshtml
TomHanmer29: Yeah this was missed out I think. Its different on the other property pages

web/Views/Property/PropertyInputForm/TenantPreferences.cshtml
TomHanmer29: The line above says "on universal credit" and affects acceptsCredit, it makes no sense to have the on universal credit bit here any more

BricksAndHearts.UnitTests/ServiceTests/CsvImport/CsvImportServiceTests.cs
```diff
@@ -26,6 +24,8 @@ public CsvImportServiceTests(TestDatabaseFixture fixture)
         MessageHandler = A.Fake<HttpMessageHandler>();
         HttpClient = new HttpClient(MessageHandler);
     }
+    private TestDatabaseFixture Fixture { get; }
+    private ILogger<CsvImportService> Logger { get; }
```
emilypatsko, line 28: Is there a reason these have moved down here?

web/Migrations/20220825144120_AddHasGuarantorAndChangeEETInTenants.cs
```diff
@@ -0,0 +1,35 @@
...
+{
+    public partial class AddHasGuarantorAndChangeEETInTenants : Migration
+    {
+        protected override void Up(MigrationBuilder migrationBuilder)
+        {
+            migrationBuilder.RenameColumn(
+                name: "NotInEET",
+                table: "Tenant",
+                newName: "InEET");
+
```
emilypatsko, line 15: I think technically to preserve the 'trueness' of our data, we'll want to negate any existing data in the column by adding a line to execute some SQL, like:
```
Sql("UPDATE Tenant SET NotInEET = 1- NotInEET;");
```
@oskarwilliams378 does this look right to you?

emilypatsko: And we'd need to do the same again in the `Down` method below

web/Services/TenantService.cs
```diff
@@ -64,12 +64,14 @@ private IQueryable<TenantDbModel> GetFilteredTenantQuery(HousingRequirementModel
...
-                                          && (filters.AcceptsNotInEET == null || t.NotInEET == filters.AcceptsNotInEET)
+                                          && (filters.AcceptsNotInEET == null || t.InEET == !filters.AcceptsNotInEET)
                                           && (filters.AcceptsCredit == null
                                               || t.UniversalCredit == filters.AcceptsCredit)
                                           && (filters.AcceptsBenefits == null
                                               || t.HousingBenefits == filters.AcceptsBenefits)
-                                          && (filters.AcceptsUnder35 == null || t.Under35 == filters.AcceptsUnder35));
+                                          && (filters.AcceptsUnder35 == null || t.Under35 == filters.AcceptsUnder35)
+                                          && (filters.AcceptsWithoutGuarantor == null
+                                              || t.HasGuarantor == filters.AcceptsWithoutGuarantor));
```
emilypatsko, line 74: I think that should be `t.HasGuarantor == !filters.AcceptsWithoutGuarantor`, shouldn't it?
e.g. if `HasGuarantor` is false, the tenant doesn't have a guarantor, so we'd want `AcceptsWithoutGuarantor` to be true

web/Migrations/20220825144120_AddHasGuarantorAndChangeEETInTenants.cs
```diff
@@ -0,0 +1,35 @@
...
+#nullable disable
+
+namespace BricksAndHearts.Migrations
+{
+    public partial class AddHasGuarantorAndChangeEETInTenants : Migration
+    {
+        protected override void Up(MigrationBuilder migrationBuilder)
+        {
+            migrationBuilder.RenameColumn(
```
oskarwilliams378, line 11: You can add something into the migration here to make the db invert all existing true and false values :)

web/Services/TenantService.cs
oskarwilliams378, line 74: should this be inverted

web/Views/Admin/_MatchListTable.cshtml
```diff
@@ -58,22 +58,26 @@
...
-                {
-                    propertyLink = Model.LandlordAndProperty.Property!.PublicViewLink,
-                    tenantEmail = tenant.Email,
-                    currentPropertyId = Model.LandlordAndProperty.Property.PropertyId
-                },
-                    FormMethod.Post))
-                {
-                    <button type="submit" class="btn btn-thin btn-outline-primary mx-3">Match</button>
-                }
+                    @using (Html.BeginForm("SendMatchLinkEmail", "Tenant", new
```
oskarwilliams378, line 71: I'd un indent this

web/Views/Property/PropertyInputForm/PropertyAccepts.cshtml
```diff
@@ -45,7 +45,7 @@
     <div class="form-check">
         <input type="checkbox" asp-for="AcceptsBenefits" class="form-check-input"/>
         <label asp-for="AcceptsBenefits" class="form-check-label">
-            on Universal Credit or housing benefits?
+            on Housing Benefits?
```
emilypatsko, line 48: Could you do me a favour and add a question mark to the one above (`on Universal Credit?`)

web/wwwroot/TenantImportCSV_Template.csv
```diff
@@ -1,2 +1,2 @@
-﻿Name,Email,Phone,Postcode,Type,HasPet,NotInEET,UniversalCredit,HousingBenefits,Under35
-Example Person,persona@gmail.com,01304 676838,cb39aj,Single,TRUE,TRUE,TRUE,TRUE,FALSE
+﻿Name,Email,Phone,Postcode,Type,HasPet,InEET,UniversalCredit,HousingBenefits,Under35,HasGuarantor
+Example Person,persona@gmail.com,01304 676838,cb39aj,Single,TRUE,TRUE,TRUE,TRUE,FALSE,TRUE
```
oskarwilliams378, line 2: Do you want to flip the InEET True False?

oskarwilliams378 (changes requested):

web/Views/Shared/Details/_PropertyAccepts.cshtml
```diff
@@ -1,58 +1,58 @@
...
+                <i class="bi me-1 @GetPreferenceIcon(Model.Property.LandlordRequirements.AcceptsBenefits)"></i>
+                On housing benefits
+            </div>
+            <div class="@GetPreferenceFont(Model.Property.LandlordRequirements.AcceptsNotInEET)">
+                <i class="bi me-1 @GetPreferenceIcon(Model.Property.LandlordRequirements.AcceptsNotInEET)"></i>
+                Not in EET (Employment, Education, or Training)
+            </div>
+            <div class="@GetPreferenceFont(Model.Property.LandlordRequirements.AcceptsWithoutGuarantor)">
+                <i class="bi me-1 @GetPreferenceIcon(Model.Property.LandlordRequirements.AcceptsWithoutGuarantor)"></i>
+                Accepts without guarantor
```
emilypatsko, line 54: I think this should just say 'without guarantor', because these all come below the 'Property accepts' heading, and it would read weirdly if 'accepts' was there twice :)

emilypatsko (changes requested):

web/Migrations/20220825144120_AddHasGuarantorAndChangeEETInTenants.cs
oskarwilliams378: I'd possibly not do it this way (mathematically), but do it with
```UPDATE Tenant SET InEET = NOT InEET```
I'd also do it with the new column names after we update the column name

File: web/Migrations/20220825144120_AddHasGuarantorAndChangeEETInTenants.cs (hunk 10 above)
Comment (emilypatsko) [reply web/Migrations/20220825144120_AddHasGuarantorAndChangeEETInTenants.cs:11]: Moving my comment from below: 
I think technically to preserve the 'trueness' of our data, we'll want to negate any existing data in the column by adding a line to execute some SQL, like:
```
Sql("UPDATE Tenant SET NotInEET = 1- NotInEET;");
```
@oskarwilliams378 does this look right to you?

Comment (emilypatsko) [reply web/Migrations/20220825144120_AddHasGuarantorAndChangeEETInTenants.cs:11]: (and we'd want to do the same in the `Down` method)

File: web/Migrations/20220825144120_AddHasGuarantorAndChangeEETInTenants.cs (hunk 8 above)
Comment (emilypatsko) [reply web/Migrations/20220825144120_AddHasGuarantorAndChangeEETInTenants.cs:15]: Two things:
- I meant to use the new column name but had tested my SQL locally first and hadn't renamed the column, forgot to change it in my comment oops
- I _wanted_ to do it this way, but MSSQL didn't like it :( and I'm not sure why. I've updated my above comment to what I think it should be @oskarwilliams378

File: web/Views/Admin/TenantMatchList.cshtml [hunk 15]
```diff
@@ -4,5 +4,7 @@
     ViewData["Title"] = "Tenant List";
 }
 
+<h1 class="font-h1">Selected property</h1>
 <partial name="~/Views/Admin/_MatchListPropertyViewPartial.cshtml" model="@Model.LandlordAndProperty"/>
+<h1 class="font-h1">Matching tenants</h1>
```
emilypatsko, line 9: My bad, I should have mentioned this - you should only ever have one `<h1>` on a page! If you want them to both _look_ like `<h1>`s, you should apply the appropriate bootstrap class to them both :)

web/Views/Shared/Details/_PropertyAccepts.cshtml
```diff
@@ -35,23 +35,23 @@
...
             <div class="@GetPreferenceFont(Model.Property.LandlordRequirements.AcceptsCredit)">
                 <i class="bi me-1 @GetPreferenceIcon(Model.Property.LandlordRequirements.AcceptsCredit)"></i>
                 On Universal Credit
             </div>
-            <div class="@GetPreferenceFont(!Model.Property.LandlordRequirements.AcceptsNotInEET)">
-                <i class="bi me-1 @GetPreferenceIcon(!Model.Property.LandlordRequirements.AcceptsNotInEET)"></i>
-                In EET (Employment, Education, or Training)
+            <div class="@GetPreferenceFont(Model.Property.LandlordRequirements.AcceptsBenefits)">
+                <i class="bi me-1 @GetPreferenceIcon(Model.Property.LandlordRequirements.AcceptsBenefits)"></i>
+                On Housing Benefits
```
emilypatsko, line 46: "On Housing Benefits" doesn't need capitalising :) just "On housing benefits"

emilypatsko (changes requested):
Basically done, just 2 v small comments

emilypatsko (approved):

